PR #63, opened 2020-11-16
AB#328 : Add /recover API endpoint

Nirusu:
Short overview:
- Add /recover API endpoint to set a new encryption key, in case the sealed state cannot be decrypted successfully with the key stored on disk
- When in "recovery" mode, the user can either upload a new valid encryption key (raw), or upload a new manifest 
- For how to use the recovery features, see [docs/recovery.md](https://github.com/edgelesssys/marblerun/blob/recovery_new/docs/recovery.md)
- When, instead of recovering the old key, a new manifest is set, we generate a new key automatically as the old one is likely botched and unusable.

Open questions:
- Is the way loadState(dnsName) is initiated by the recovery API endpoint okay? loadState takes dnsName as a parameter in case the it has to generate a new certificate. Right now, when calling from the /recover endpoint, it will use `nil` as parameter and inside loadState, we check for nil to see if we're stuck in recovery.
- Does it make sense to save the old recovery key? I thought directly overwriting the old key permanently seems a bit aggressive, but does it really make any sense to store the old key?

### DIFF
--- a/coordinator/core/clientapi.go
+++ b/coordinator/core/clientapi.go
@@ -25,7 +25,7 @@ type ClientCore interface {
 	GetCertQuote(ctx context.Context) (cert string, certQuote []byte, err error)
 	GetManifestSignature(ctx context.Context) (manifestSignature []byte)
 	GetStatus(ctx context.Context) (statusCode int, status string, err error)
-	SetEncryptionKey(ctx context.Context, encryptionKey []byte) error
+	Recover(ctx context.Context, encryptionKey []byte) error
 }
 
 // SetManifest sets the manifest, once and for all
@@ -116,8 +116,8 @@ func (c *Core) GetManifestSignature(ctx context.Context) []byte {
 	return hash[:]
 }
 
-// SetEncryptionKey sets an encryption key (ideally decrypted from the recovery data) and tries to unseal and load a saved state again.
-func (c *Core) SetEncryptionKey(ctx context.Context, encryptionKey []byte) error {
+// Recover sets an encryption key (ideally decrypted from the recovery data) and tries to unseal and load a saved state again.
+func (c *Core) Recover(ctx context.Context, encryptionKey []byte) error {
 	defer c.mux.Unlock()
 	if err := c.requireState(stateRecovery); err != nil {
 		return err
--- a/coordinator/core/clientapi.go
+++ b/coordinator/core/clientapi.go
@@ -25,6 +25,7 @@ type ClientCore interface {
 	GetCertQuote(ctx context.Context) (cert string, certQuote []byte, err error)
 	GetManifestSignature(ctx context.Context) (manifestSignature []byte)
 	GetStatus(ctx context.Context) (status string, err error)
+	SetEncryptionKey(ctx context.Context, encryptionKey []byte) error
 }
 
 // SetManifest sets the manifest, once and for all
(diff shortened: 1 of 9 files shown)


Nirusu:
Added the callback for core.go & server.go (but not the marbles itself) & rebased to new master. Hope I didn't miss anything...?

coordinator/core/core.go
```diff
@@ -259,10 +266,22 @@ func (c *Core) generateCert(dnsNames []string) (*x509.Certificate, *ecdsa.Privat
 		return nil, nil, err
 	}
 
-	c.advanceState()
 	return cert, privk, nil
 }
 
+func (c *Core) generateQuote() ([]byte, error) {
```
thomasten, line 272:
```suggestion
func (c *Core) generateQuote() []byte {
```

test/integration_test.go
```diff
@@ -238,6 +244,202 @@ func TestClientAPI(t *testing.T) {
...
+
+	// Test with certificate
+	log.Println("Verifying certificate after recovery, without a restart.")
+	pool := x509.NewCertPool()
+	require.True(pool.AppendCertsFromPEM([]byte(cert)))
+	client = http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{RootCAs: pool}}}
+	clientAPIURL.Path = "status"
+	resp, err = client.Get(clientAPIURL.String())
+	require.NoError(err)
+	require.Equal(http.StatusOK, resp.StatusCode)
```
thomasten, line 333:
```suggestion
	require.NoError(err)
	resp.Body.Close()
	require.Equal(http.StatusOK, resp.StatusCode)
```

test/integration_test.go
```diff
@@ -238,6 +244,202 @@ func TestClientAPI(t *testing.T) {
...
+	assert.EqualValues(3, gjson.Get(statusResponse, "Code").Int(), "Server is in wrong status after recovery.")
+
+	// test with certificate
+	log.Println("Verifying certificate after restart.")
+	pool = x509.NewCertPool()
+	require.True(pool.AppendCertsFromPEM([]byte(cert)))
+	client = http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{RootCAs: pool}}}
+	clientAPIURL.Path = "status"
+	resp, err = client.Get(clientAPIURL.String())
+	require.NoError(err)
```
thomasten, line 359:
```suggestion
	resp, err = client.Get(clientAPIURL.String())
	require.NoError(err)
	resp.Body.Close()
```

thomasten (changes requested):

coordinator/core/clientapi.go
```diff
@@ -108,12 +115,33 @@ func (c *Core) GetManifestSignature(ctx context.Context) []byte {
...
+
+	c.cert = cert
+	c.privk = privk
+
+	return nil
+}
+
+// GetStatus returns status information about the state of the mesh.
+func (c *Core) GetStatus(ctx context.Context) (statusCode int, status string, err error) {
 	if err != nil {
```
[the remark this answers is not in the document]
Nirusu: What do you mean with "block"?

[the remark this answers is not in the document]
Nirusu: Oh wait, I see. Was kinda confusing with what GitHub showed here.

thomasten (approved):

PR #70, opened 2020-11-23
AB#361 Change manifest placeholders to Go templates

Nirusu:
First attempt of using Go templates for secrets.

This is not the final design for storing & accessing secrets yet. So far, it's only a replacement for the $$ placeholder strings in the manifest so far.

Tests should be covered with the existing TestActivate test in marbleapi_test.go.

### DIFF
--- a/coordinator/core/marbleapi.go
+++ b/coordinator/core/marbleapi.go
@@ -7,15 +7,14 @@
 package core
 
 import (
+	"bytes"
 	"context"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/x509"
-	"encoding/hex"
-	"encoding/pem"
 	"math"
-	"strings"
+	"text/template"
 	"time"
 
 	"github.com/edgelesssys/marblerun/coordinator/rpc"
@@ -25,6 +24,12 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
+
+type marbleSecretStruct struct {
+	RootCA     Secret
+	MarbleCert Secret
+	SealKey    Secret
+}
 
 // Activate implements the MarbleAPI function to authenticate a marble (implements the MarbleServer interface)
 //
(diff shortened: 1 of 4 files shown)


coordinator/core/marbleapi.go
```diff
@@ -81,12 +90,38 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
...
+
+	marbleSecretsWrapped := marbleSecretsWrapper{
+		Marblerun: marbleSecrets,
+	}
+
 	marble := c.manifest.Marbles[req.GetMarbleType()] // existence has been checked in verifyManifestRequirement
-	params := customizeParameters(marble.Parameters, pemRootCA, pemMarbleCert, pemMarbleKey, strSealKey)
+	params, err := customizeParameters(marble.Parameters, marbleSecretsWrapped)
+	if err != nil {
+		c.zaplogger.Error("Could not customize parameters.", zap.Error(err))
```
thomasten, line 123:
```suggestion
		c.zaplogger.Error("Could not customize parameters.", zap.Error(err))
		return nil, err
```

coordinator/core/marbleapi.go
```diff
@@ -81,12 +90,38 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
...
+		Name:    "sealKey",
+		Public:  sealKey,
+		Private: sealKey,
+	}
+
+	marbleSecrets := marbleSecretsStruct{
+		RootCA:     pemRootCAObject,
+		MarbleCert: pemMarbleCertObject,
+		SealKey:    strSealKeyObject,
+	}
```
thomasten, line 114:
```suggestion
	marbleSecrets := marbleSecretsStruct{
		RootCA:     Secret{Public: c.cert.Raw},
		MarbleCert: Secret{Public: certRaw, Private: encodedPrivKey},
		SealKey:    Secret{Public: sealKey, Private: sealKey},
	}
```
I think we don't need `Name`

coordinator/core/marbleapi.go
```diff
@@ -81,12 +90,38 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
...
+
+	marbleSecretsWrapped := marbleSecretsWrapper{
```
thomasten, line 116: I'd move that into customizeParameters

coordinator/core/marbleapi.go
```diff
@@ -177,26 +212,42 @@ func (c *Core) generateCertFromCSR(csrReq []byte, pubk ecdsa.PublicKey, marbleTy
...
+	var templateResult bytes.Buffer
 	// replace placeholders in files
-	r := strings.NewReplacer(rootCAPlaceholder, string(rootCA),
-		marbleCertPlaceholder, string(marbleCert),
-		marbleKeyPlaceholder, string(marbleKey),
-		sealKeyPlaceholder, sealKey)
 	for path, data := range params.Files {
-		newData := r.Replace(data)
-		customParams.Files[path] = newData
+		tpl, err := template.New("data").Funcs(manifestTemplateFuncMap).Parse(data)
```
thomasten, line 225: refactor the code in the two loops into a common function

test/manifests.go
```diff
@@ -55,10 +55,10 @@ const ManifestJSON string = `{
 				},
 				"Env": {
 					"IS_FIRST": "true",
-					"ROOT_CA": "$$root_ca",
-					"SEAL_KEY": "$$seal_key",
-					"MARBLE_CERT": "$$marble_cert",
-					"MARBLE_KEY": "$$marble_key"
+					"ROOT_CA": "{{ pem .Marblerun.RootCA.Public }}",
```
thomasten, line 58: also update the example in BUILD.md

coordinator/core/marbleapi.go
```diff
@@ -177,26 +212,42 @@ func (c *Core) generateCertFromCSR(csrReq []byte, pubk ecdsa.PublicKey, marbleTy
 }
 
 // customizeParameters replaces the placeholders in the manifest's parameters with the actual values
-func customizeParameters(params *rpc.Parameters, rootCA []byte, marbleCert []byte, marbleKey []byte, sealKey string) *rpc.Parameters {
+func customizeParameters(params *rpc.Parameters, marbleSecretsWrapped marbleSecretsWrapper) (*rpc.Parameters, error) {
```
thomasten, line 215: add a unit test for this function

Nirusu: `TestActivate` does already test the reserved secrets and will fail if e.g. the generated rootCA PEM is invalid and cannot be decoded correctly. 

Considering support for other, non-reserved secrets is coming in an upcoming PR, I'd suggest we create a separate unit test then which will then test the general implementation of the templates. But for the specific case of the "reserved" secrets in this PR, it should already be covered.

thomasten (approved):
Nice job!

thomasten (approved):

PR #71, opened 2020-11-24
AB#351 & AB#362 : Secret store in core & ability to generate symmetric secrets

Nirusu:
(Based on #70 , check the base branch before merging!)

Let's get this ready for a review. Based on the other PR, this one adds storing secrets in the client's core, specifying secrets in a manifest and injecting them secrets into a marble.

This PR only covers shared secrets of the type "raw" so far.

One question that comes to my mind is that if we should enforce some kind of boundaries, limits or some additional sanity check? Right now, anything can be defined e.g. when it comes to size. Is this desirable, or do we want some stricter limits?

### DIFF
--- a/test/manifests.go
+++ b/test/manifests.go
@@ -58,7 +58,10 @@ const ManifestJSON string = `{
 					"ROOT_CA": "{{ pem .Marblerun.RootCA.Public }}",
 					"SEAL_KEY": "{{ hex .Marblerun.SealKey }}",
 					"MARBLE_CERT": "{{ pem .Marblerun.MarbleCert.Public }}",
-					"MARBLE_KEY": "{{ pem .Marblerun.MarbleCert.Private }}"
+					"MARBLE_KEY": "{{ pem .Marblerun.MarbleCert.Private }}",
+					"TEST_SECRET_RAW": "{{ raw .Secrets.testsecret_raw }}",
+					"TEST_SECRET_HEX": "{{ hex .Secrets.testsecret_raw }}",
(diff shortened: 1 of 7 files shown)


test/manifests.go
```diff
@@ -173,7 +182,8 @@ var IntegrationManifestJSON string = `{
 					"ROOT_CA": "{{ pem .Marblerun.RootCA.Public }}",
 					"SEAL_KEY": "{{ hex .Marblerun.SealKey }}",
 					"MARBLE_CERT": "{{ pem .Marblerun.MarbleCert.Public }}",
-					"MARBLE_KEY": "{{ pem .Marblerun.MarbleCert.Private }}"
+					"MARBLE_KEY": "{{ pem .Marblerun.MarbleCert.Private }}",
+					"TEST_SECRET_HEX": "{{ hex .Secrets.testsecret_raw }}"
```
thomasten, line 186:
```suggestion
					"MARBLE_KEY": "{{ pem .Marblerun.MarbleCert.Private }}"
```
I think we don't need this in the integration test. We should limit the integration test to parts that can't be covered with unit tests.

test/manifests.go
```diff
@@ -212,6 +222,12 @@ var IntegrationManifestJSON string = `{
 	"Clients": {
 		"owner": [9,9,9]
 	},
+	"Secrets": {
+		"testsecret_raw": {
+			"size": 16,
+			"type": "raw"
+		}
+	},
```
thomasten, line 230:
```suggestion
```

thomasten (changes requested):

thomasten (approved):

PR #76, opened 2020-12-07
AB#393 : Marble unique secrets

Nirusu:
Can be reviewed together with the documentation at edgelesssys/marblerun.sh#6

### DIFF
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -335,9 +335,8 @@ func (c *Core) generateSecrets(ctx context.Context, secrets map[string]Secret) (
 			}
 
 			var generatedValue []byte
-
 			// If a secret is shared, we generate a completely random key. If a secret is constrained to a marble, we derive a key from the core's private key.
-			if secret.Shared == true {
+			if secret.Shared {
 				generatedValue = make([]byte, secret.Size/8)
 				_, err := rand.Read(generatedValue)
 				if err != nil {
@@ -345,11 +344,9 @@ func (c *Core) generateSecrets(ctx context.Context, secrets map[string]Secret) (
 				}
 			} else {
 				salt := secret.UUID.String() + name
-				secretKeyDerive, err := x509.MarshalPKCS8PrivateKey(c.privk)
-				if err != nil {
-					return nil, err
-				}
-				generatedValue, err = util.DeriveKey(secretKeyDerive, []byte(salt), (secret.Size / 8))
+				secretKeyDerive := c.privk.D.Bytes()
+				var err error
+				generatedValue, err = util.DeriveKey(secretKeyDerive, []byte(salt), secret.Size/8)
 				if err != nil {
 					return nil, err
 				}
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/edgelesssys/marblerun/coordinator/quote"
 	"github.com/edgelesssys/marblerun/util"
+	"github.com/google/uuid"
 	"go.uber.org/zap"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/peer"
@@ -319,12 +320,20 @@ func (c *Core) getStatus(ctx context.Context) (int, string, error) {
 	return int(c.state), status, nil
 }
 
-func (c *Core) generateSecrets(ctx context.Context, secrets map[string]Secret) (map[string]Secret, error) {
+func (c *Core) generateSecrets(ctx context.Context, secrets map[string]Secret, id uuid.UUID) (map[string]Secret, error) {
 	// Create a new map so we do not overwrite the entries in the manifest
 	newSecrets := make(map[string]Secret)
 
 	// Generate secrets
 	for name, secret := range secrets {
+
+		// Skip secrets from wrong context
+		if secret.Shared && id != uuid.Nil {
+			continue
+		} else if !secret.Shared && id == uuid.Nil {
+			continue
+		}
+
 		c.zaplogger.Info("generating secret", zap.String("name", name), zap.String("type", secret.Type), zap.Uint("size", secret.Size))
 		switch secret.Type {
 		// Raw = Symmetric Key
@@ -343,7 +352,7 @@ func (c *Core) generateSecrets(ctx context.Context, secrets map[string]Secret) (
 					return nil, err
 				}
 			} else {
-				salt := secret.UUID.String() + name
+				salt := id.String() + name
 				secretKeyDerive := c.privk.D.Bytes()
 				var err error
 				generatedValue, err = util.DeriveKey(secretKeyDerive, []byte(salt), secret.Size/8)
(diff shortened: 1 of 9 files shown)


coordinator/core/marbleapi.go
```diff
@@ -107,8 +107,29 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
 		SealKey:    Secret{Public: sealKey, Private: sealKey},
 	}
 
+	// Generate unique (= per marble) secrets
+	uniqueSecrets := make(map[string]Secret)
+	for k, v := range c.manifest.Secrets {
+		if v.Shared == false {
+			v.UUID = marbleUUID
```
thomasten, line 114: As every secret gets the same uuid here, I'd propose to not have this as a field in the struct, but pass it as string to generateSecrets. Then you can also remove this filtering loop and the one in SetManifest and instead skip secrets in generateSecrets based on whether uuid is empty or not.

coordinator/core/marbleapi.go
```diff
@@ -107,8 +107,29 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
...
+	if err != nil {
+		c.zaplogger.Error("Could not generate specified secrets for the given manifest.", zap.Error(err))
+		return nil, err
+	}
+
+	// Union unique secrets with shared secrets
+	for k, v := range c.secrets {
+		if v.Shared == true {
+			secrets[k] = v
+		}
```
thomasten, line 128:
```suggestion
		secrets[k] = v
```
always true

coordinator/core/marbleapi.go
```diff
@@ -107,8 +107,29 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
 		SealKey:    Secret{Public: sealKey, Private: sealKey},
```
thomasten, line 107: Not directly related to this PR, but the function is a bit long. Can you refactor the code from
privk, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
till
authSecrets := reservedSecrets{
into a new function?

coordinator/core/marbleapi_test.go
```diff
@@ -32,6 +32,12 @@ import (
 	"google.golang.org/grpc/peer"
 )
 
+// Used to check if shared secrets are the same and non-shared secrets are indeed unique
+var backendFirstSharedCert x509.Certificate
```
thomasten, line 36: Pls move this into the marbleSpawner struct. Avoid global variables.

coordinator/core/marbleapi_test.go
```diff
@@ -199,18 +210,80 @@ func (ms *marbleSpawner) newMarble(marbleType string, infraName string, shouldSu
...
+			backendFirstUniqueCert = *secretCertNonShared
+		}
+		ms.mutex.Unlock()
+
+		// Check if our certificate does actually expire 7 days, as specified, after it was generated
+		expectedNotBefore = secretCertNonShared.NotAfter.AddDate(0, 0, -7)
+		ms.assert.EqualValues(expectedNotBefore, secretCertNonShared.NotBefore)
+
+	} else if marbleType == "backend_other" {
+		// Validate generated shared secret certificate
```
thomasten, line 256: This block is duplicated 4 times. Try to refactor.

thomasten (changes requested):

coordinator/core/core.go
```diff
@@ -334,10 +334,25 @@ func (c *Core) generateSecrets(ctx context.Context, secrets map[string]Secret) (
...
+			// If a secret is shared, we generate a completely random key. If a secret is constrained to a marble, we derive a key from the core's private key.
+			if secret.Shared == true {
+				generatedValue = make([]byte, secret.Size/8)
+				_, err := rand.Read(generatedValue)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				salt := secret.UUID.String() + name
+				secretKeyDerive, err := x509.MarshalPKCS8PrivateKey(c.privk)
```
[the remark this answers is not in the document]
Nirusu: Nope, was just used to work on the marshalled versions from the previous PR that I immediately used it this way :D 

I'll change it.

coordinator/core/marbleapi_test.go
```diff
@@ -295,6 +237,22 @@ func (ms *marbleSpawner) newMarbleAsync(marbleType string, infraName string, sho
 	}()
 }
 
+func (ms *marbleSpawner) verifyCertificateFromEnvironment(envName string, params *rpc.Parameters, opts x509.VerifyOptions) x509.Certificate {
```
Nirusu, line 240: Shoud this return an error, panic or is the current way (assert inside the function) okay?

coordinator/core/core.go
```diff
@@ -319,12 +320,20 @@ func (c *Core) getStatus(ctx context.Context) (int, string, error) {
...
 
 	// Generate secrets
 	for name, secret := range secrets {
+
+		// Skip secrets from wrong context
+		if secret.Shared && id != uuid.Nil {
+			continue
+		} else if !secret.Shared && id == uuid.Nil {
+			continue
+		}
```
thomasten, line 335:
```suggestion
		if secret.Shared != (id == uuid.Nil) {
			continue
		}
```
I find this easier to read, but if you disagree, you can keep it as is.

coordinator/core/marbleapi_test.go
```diff
@@ -222,6 +237,22 @@ func (ms *marbleSpawner) newMarbleAsync(marbleType string, infraName string, sho
...
+	p, _ := pem.Decode([]byte(params.Env[envName]))
+	ms.assert.NotNil(p)
+	certificate, err := x509.ParseCertificate(p.Bytes)
+	ms.assert.NotNil(p)
+	ms.assert.NoError(err)
```
thomasten, line 245:
```suggestion
	ms.require.NotNil(p)
	certificate, err := x509.ParseCertificate(p.Bytes)
	ms.require.NoError(err)
```

thomasten (changes requested):

coordinator/core/marbleapi_test.go
thomasten: okay, but see comment regarding `require` below

coordinator/core/core.go
Nirusu: That's fine, I'll take it

coordinator/core/marbleapi_test.go
Nirusu: I left the root CA check & NotBefore check as asserts, is this okay? Not quite sure when to use assert and when to use require...

thomasten: Yep. require stops the test whereas assert just marks it as failed. So you use require if the rest of the test depends on the result. E.g., if p is nil, there would be a nil dereference.

thomasten (approved):

PR #80, opened 2020-12-14
AB#431 : Use new sealing features from ertcrypto

Nirusu:
The encryption key is now sealed with the CPU sealing key using the new ertcrypto features. Sealing the state data, recovery etc. works the same way as before.


thomasten (changes requested):
Integration test with -noenclave is broken. We probably need another kind of mock sealer that just writes the plaintext to disk.

Nirusu:
Done. 
Is an unit test needed for this one? If yes, should we just go full-length and completely replace the MockSealer with the new NoEnclaveSealer?

thomasten:
> Is an unit test needed for this one? If yes, should we just go full-length and completely replace the MockSealer with the new NoEnclaveSealer?

No, not needed. The Sealer interface exists mainly to make the core unit testable. So Sealer implementations are supposed to be integration-tested only.

thomasten (approved):

PR #83, opened 2020-12-21
AB#457 : Check combination of package properties

Nirusu:
Note: This does not cover ERTValidator (it's still reports that  it would be compliant package), the coordinator does the check when setting the manifest. This is so we can handle debug/non-debug mode differently without introducing larger changes.

### DIFF
--- a/coordinator/core/clientapi_test.go
+++ b/coordinator/core/clientapi_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/edgelesssys/marblerun/test"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func mustSetup() (*Core, *Manifest) {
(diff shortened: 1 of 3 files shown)


coordinator/core/clientapi_test.go
```diff
@@ -73,9 +75,30 @@ func TestSetManifest(t *testing.T) {
 		break
 	}
 	modRawManifest, err := json.Marshal(manifest)
-	assert.NoError(err)
+	require.NoError(err)
 	_, err = c.SetManifest(context.TODO(), modRawManifest)
 	assert.Equal("manifest does not contain marble package foo", err.Error())
+
+	// Try setting manifest with UniqueID + other value set, no debug mode (this should fail)
```
thomasten, line 82: In addition to what you have implemented and test here, we also want to enforce that if UniqueID is not set, the other 3 values must all be set. So setting, e.g., only SignerID and ProductID is invalid.
When you add this functionality and more tests, it is probably a good idea to move these test into its own Test... func.

thomasten (approved):

PR #85, opened 2021-01-12
AB#494 & AB#496 : Allow updating the required SecurityVersion of a package

Nirusu:
Short summary of changes:

- New section `Admins` in manifest, allowing to specify **user-defined** certificates (which get saved to the core)
- Introduction of an 'update manifest', which allows to specify a new, higher SecurityLevel than set in the original manifest
- This 'update manifest' does not overwrite the original manifest, but is saved additionally in the core.
- `verifyManifestRequirement` in `marbleapi.go` overwrites the SecurityVersion for each specified package in the update manifest. This should replace the SecurityLevel right in time before it passes the Validator.
- The update manifest can be set over the '/update' API endpoint, when one of the client certificates from `Admins` was used (and if none was set, this API endpoint stays locked)
- The certificates are **not fully verified**. We only check if the supplied one is equal to one of the defined ones in `Admins`, and Go's TLS stack should ensure that the user calling the API endpoint does own the private key to its own certificate. Expiration and trust anchors are not checked here, we basically just abuse TLS client certificates here for a public/private key check and do not care about the rest of the X.509 certificate stuff. Guess we need to check carefully here if this provides sufficient security.

### DIFF
--- a/coordinator/core/clientapi.go
+++ b/coordinator/core/clientapi.go
@@ -82,25 +82,17 @@ func (c *Core) SetManifest(ctx context.Context, rawManifest []byte) ([]byte, err
 		return nil, err
 	}
 
+	// Parse X.509 admin certificates from manifest
+	adminCerts, err := generateAdminCertsFromManifest(manifest.Admins)
+	if err != nil {
+		c.zaplogger.Error("Could not parse specified admin client certificate from supplied manifest", zap.Error(err))
+		return nil, err
+	}
+
 	c.manifest = manifest
 	c.rawManifest = rawManifest
 	c.secrets = secrets
-
-	// Parse & write X.509 admin certificates to core
-	for _, value := range manifest.Admins {
-		block, _ := pem.Decode([]byte(value))
-		if err != nil {
-			c.zaplogger.Error("Could not decode specified admin client certificate", zap.Error(err))
-			return nil, err
-		}
-		cert, err := x509.ParseCertificate(block.Bytes)
-		if err != nil {
-			c.zaplogger.Error("Could not parse specified admin client certificate", zap.Error(err))
-			return nil, err
-		}
-
-		c.adminCerts = append(c.adminCerts, cert)
-	}
+	c.adminCerts = adminCerts
 
 	c.advanceState(stateAcceptingMarbles)
 	encryptionKey, err := c.sealState()
@@ -163,7 +155,6 @@ func (c *Core) Recover(ctx context.Context, encryptionKey []byte) error {
 
 	c.cert = cert
 	c.privk = privk
-
 	c.quote = c.generateQuote()
 
 	return nil
@@ -176,20 +167,17 @@ func (c *Core) GetStatus(ctx context.Context) (statusCode int, status string, er
 
 // VerifyAdmin checks if a given client certificate matches the admin certificates specified in the manifest
 func (c *Core) VerifyAdmin(ctx context.Context, clientCerts []*x509.Certificate) bool {
-	matchedCertificate := false
-
 	// Check if a supplied client cert matches the supplied ones from the manifest stored in the core
 	// NOTE: We do not use the "correct" X.509 verify here since we do not really care about expiration and chain verification here.
 	for _, suppliedCert := range clientCerts {
 		for _, knownCert := range c.adminCerts {
 			if suppliedCert.Equal(knownCert) {
-				matchedCertificate = true
-				break
+				return true
 			}
 		}
 	}
 
-	return matchedCertificate
+	return false
 }
 
 // UpdateManifest allows to update certain package parameters, supplied via a JSON manifest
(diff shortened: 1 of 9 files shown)


thomasten (changes requested):

coordinator/core/manifest.go
```diff
@@ -49,6 +51,11 @@ type Marble struct {
 	Parameters *rpc.Parameters
 }
 
+// UpdateManifest describes a update manifest which can be used to update certain package properties
+type UpdateManifest struct {
```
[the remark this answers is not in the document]
Nirusu: We can, and I changed it in the last commit. I'll just use a different check function for the update case, and other values should automatically be ignored for now.

coordinator/core/clientapi.go
```diff
@@ -80,9 +82,27 @@ func (c *Core) SetManifest(ctx context.Context, rawManifest []byte) ([]byte, err
 		return nil, err
 	}
 
+	// Parse & write X.509 admin certificates to core
+	adminCerts := make([]*x509.Certificate, len(manifest.Admins))
```
thomasten, line 86:
```suggestion
	adminCerts := make([]*x509.Certificate)
```
or
```suggestion
	adminCerts := make([]*x509.Certificate, 0, len(manifest.Admins))
```

coordinator/core/core.go
```diff
@@ -168,41 +175,67 @@ func (c *Core) GetTLSCertificate(clientHello *tls.ClientHelloInfo) (*tls.Certifi
...
+		}
+		c.rawUpdateManifest = loadedState.RawUpdateManifest
+	}
+
 	c.state = loadedState.State
 	c.activations = loadedState.Activations
 	c.secrets = loadedState.Secrets
-	return cert, privk, err
+
+	// Parse & write X.509 admin certificates from sealed state
```
thomasten, line 221: refactor into function and also use it in SetManifest

coordinator/core/core.go
```diff
@@ -168,41 +175,67 @@ func (c *Core) GetTLSCertificate(clientHello *tls.ClientHelloInfo) (*tls.Certifi
...
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			c.zaplogger.Error("Could not parse specified admin client certificate from sealed state", zap.Error(err))
+			return nil, nil, nil, err
+		}
+
+		adminCerts = append(adminCerts, cert)
+	}
+
+	return cert, adminCerts, privk, err
```
thomasten, line 238: I think we don't need to return adminCerts. Just create it before the other assignments above and then assign it with the others.

coordinator/core/manifest.go
```diff
@@ -201,6 +203,34 @@ func (m Manifest) Check(ctx context.Context, zaplogger *zap.Logger) error {
...
+	}
+
+	// Check if manifest update contains values which we normally should not update
+	for packageName, singlePackage := range m.Packages {
+		if singlePackage.Debug != false || singlePackage.UniqueID != "" || singlePackage.SignerID != "" || singlePackage.ProductID != nil {
+			return errors.New("update manifest contains unupdatable values")
+		}
+
+		// Checks based on the original manifest
+		if *singlePackage.SecurityVersion < *originalPackages[packageName].SecurityVersion {
```
thomasten, line 219: singlePackage.SecurityVersion might be nil
originalPackages[packageName] might not exist
originalPackages[packageName].SecurityVersion might be nil

coordinator/server/server_test.go
```diff
@@ -130,3 +170,31 @@ func TestConcurrent(t *testing.T) {
...
+	}
+
+	block, _ := pem.Decode(test.AdminCert)
+	adminTestCert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		panic(err)
+	}
+
+	return adminTestCert, otherTestCert
+}
```
thomasten, line 200:
```suggestion
```

thomasten (changes requested):

coordinator/core/manifest.go
```diff
@@ -201,6 +203,47 @@ func (m Manifest) Check(ctx context.Context, zaplogger *zap.Logger) error {
...
+		if singlePackage.SecurityVersion == nil {
+			return errors.New("update manifest does not specifiy a SecurityVersion to update")
+		}
+
+		// Check based on the original manifest
+		if originalPackages[packageName].SecurityVersion != nil && *singlePackage.SecurityVersion < *originalPackages[packageName].SecurityVersion {
+			return errors.New("update manifest tries to downgrade SecurityVersion of the original manifest")
+		}
+
+		// Checks if we already have an update manifest set, if it does contain the package and if it does, if it actually holds a value for SecurityVersion (which should always be the case, but let's go safe here)
```
thomasten, line 234: Also make sure that the new update manifest contains all the packages of alreadyUpdatedPackages.

thomasten (changes requested):

coordinator/core/manifest.go
```diff
@@ -201,6 +203,52 @@ func (m Manifest) Check(ctx context.Context, zaplogger *zap.Logger) error {
...
+		if alreadyUpdatedPackages != nil {
```
thomasten, line 235:
```suggestion
```
range nil is fine

thomasten (commented):
I think this is fine now. Pls clean up the commit history before we merge this (see workflow steps 5 and 6).

thomasten (approved):

PR #87, opened 2021-01-25
AB#377: Modularize recovery

Nirusu:
This PR makes the recovery module interchangeable for upcoming features such as multi-party recovery.

Changes from the already reviewed internal version:

- recoveryMap is always loaded into the core, not just when an unsealing error occurs 
- GetRecoveryData() added (for when we need to reseal the state for an update manifest)
- Properly some other minor stuff to accompany changes from other changes to Marblerun while this was in development

Leftover issues:

- The integration test can fail badly across tests when one of the subfunctions fails, as they often invoke `require` but no defer functions are set in them as we want to return the cfgs and alive processes for the calling main function. This can cause the next test function to work with an unclean coordinator, causing further test failures due to an unexpected state.

Guess we need an idea on how to cleanup a failure in this subfunction in a way which is not totally ugly, so any ideas are appreciated here.

Apart from that, this is ready for a final review.

### DIFF
--- a/coordinator/core/clientapi.go
+++ b/coordinator/core/clientapi.go
@@ -78,9 +78,9 @@ func (c *Core) SetManifest(ctx context.Context, rawManifest []byte) ([]byte, err
 	}
 
 	// Generate a new encryption key for a new manifest, as the old one might be broken
-	if err := c.sealer.GenerateNewEncryptionKey(); err != nil {
-		return nil, err
-	}
+	encryptionKey := make([]byte, 16)
+	_, err = rand.Read(encryptionKey)
+	c.sealer.SetEncryptionKey(encryptionKey)
 
 	// Parse X.509 admin certificates from manifest
 	adminCerts, err := generateAdminCertsFromManifest(manifest.Admins)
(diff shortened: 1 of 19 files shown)


util/util.go
```diff
@@ -56,3 +59,25 @@ func MustGetLocalListenerAndAddr() (net.Listener, string) {
 	}
 	return listener, localhost + port
 }
+
+// XORBytes XORs two byte slices
+func XORBytes(a, b []byte) ([]byte, error) {
```
thomasten, line 64: add the test

coordinator/core/clientapi.go
```diff
@@ -201,10 +178,36 @@ func (c *Core) UpdateManifest(ctx context.Context, rawUpdateManifest []byte) err
 	c.updateManifest = updateManifest
 	c.rawUpdateManifest = rawUpdateManifest
 	c.zaplogger.Info("An update manifest overriding package settings from the original manifest was set.")
-	_, err := c.sealState()
+
+	// Retrieve current recovery data before we seal the state again
```
thomasten, line 182: do this before setting `c.updateManifest` so that the state does not change on error

coordinator/core/clientapi.go
```diff
@@ -201,10 +178,36 @@ func (c *Core) UpdateManifest(ctx context.Context, rawUpdateManifest []byte) err
...
+		c.zaplogger.Error("Cannot reseal the state. Could not retrieve the current recovery data from the recovery module.")
+		return err
+	}
+
+	err = c.sealState(currentRecoveryData)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
thomasten, line 194:
```suggestion
	return c.sealState(currentRecoveryData)
```

thomasten (changes requested):
Thanks for the clean commits, much appreciated.

thomasten (approved):

m1ghtym0 (dismissed):

thomasten (approved):

PR #90, opened 2021-01-26
AB#495: Use Intermediate CA for Marble authentication

Nirusu:
This still lacks some good tests cases to check against the desired behavior, but I guess it can already be checked for sanity (it does pass all current tests).

Right now, the following certificates are used:

- Marbles: Intermediate CA
- Quoting: Root CA
- Secrets (non-shared symmetric-key derivation): Root CA Private Key
- Secrets (certificates): Intermediate CA

The intermediate CA uses a separate private/public key pair than the root CA. 

Shared Certificate secrets are re-generated when a new manifest is set and the old ones in the core will be overwritten (or should, there's no test for that yet ;)). Non-shared certificate secrets are regenerated when a Marble is activated as before , so theoretically nothing should've changed here in terms of general behavior.

Feedback would be appreciated :)

### DIFF
--- a/coordinator/core/clientapi.go
+++ b/coordinator/core/clientapi.go
@@ -91,11 +91,23 @@ func (c *Core) SetManifest(ctx context.Context, rawManifest []byte) (map[string]
 //
 // Returns the a remote attestation quote of its own certificate alongside this certificate that allows to verify the Coordinator's integrity and authentication for use of the ClientAPI.
 func (c *Core) GetCertQuote(ctx context.Context) (string, []byte, error) {
-	pemCert := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: c.rootCert.Raw})
-	if len(pemCert) <= 0 {
-		return "", nil, errors.New("pem.EncodeToMemory failed")
+	defer c.mux.Unlock()
+	if err := c.requireState(stateAcceptingManifest, stateAcceptingMarbles); err != nil {
+		return "", nil, err
 	}
-	strCert := string(pemCert)
+
+	pemCertRoot := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: c.rootCert.Raw})
+	if len(pemCertRoot) <= 0 {
+		return "", nil, errors.New("pem.EncodeToMemory failed for root certificate")
+	}
+
+	// Include intermediate certificate if a manifest has been set
+	pemCertIntermediate := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: c.intermediateCert.Raw})
+	if len(pemCertIntermediate) <= 0 {
+		return "", nil, errors.New("pem.EncodeToMemory failed for intermediate certificate")
+	}
+
+	strCert := string(pemCertIntermediate) + string(pemCertRoot)
 	return strCert, c.quote, nil
 }
 
--- a/coordinator/core/clientapi.go
+++ b/coordinator/core/clientapi.go
@@ -91,7 +91,7 @@ func (c *Core) SetManifest(ctx context.Context, rawManifest []byte) (map[string]
 //
 // Returns the a remote attestation quote of its own certificate alongside this certificate that allows to verify the Coordinator's integrity and authentication for use of the ClientAPI.
 func (c *Core) GetCertQuote(ctx context.Context) (string, []byte, error) {
-	pemCert := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: c.cert.Raw})
+	pemCert := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: c.rootCert.Raw})
 	if len(pemCert) <= 0 {
 		return "", nil, errors.New("pem.EncodeToMemory failed")
 	}
@@ -175,6 +175,13 @@ func (c *Core) UpdateManifest(ctx context.Context, rawUpdateManifest []byte) err
 		return err
 	}
 
+	// Generate new intermediate CA for Marble gRPC authentication
+	intermediateCert, intermediatePrivK, err := generateCert(c.rootCert.DNSNames, coordinatorIntermediateName, c.rootCert, c.rootPrivK)
+	if err != nil {
+		c.zaplogger.Error("Could not generate a new intermediate CA for Marble authentication.", zap.Error(err))
+		return err
+	}
+
 	// Retrieve current recovery data before we seal the state again
 	currentRecoveryData, err := c.recovery.GetRecoveryData()
 	if err != nil {
@@ -184,7 +191,11 @@ func (c *Core) UpdateManifest(ctx context.Context, rawUpdateManifest []byte) err
 
 	c.updateManifest = updateManifest
 	c.rawUpdateManifest = rawUpdateManifest
+	c.intermediateCert = intermediateCert
+	c.intermediatePrivK = intermediatePrivK
+
 	c.zaplogger.Info("An update manifest overriding package settings from the original manifest was set.")
+	c.zaplogger.Info("Please restart your Marbles to enforce the update.")
 
 	return c.sealState(currentRecoveryData)
 }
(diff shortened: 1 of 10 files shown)


m1ghtym0 (approved):
LGTM, we should think about returning the rootCert and intermediateCert on the client API.
The client can decide which certificate to pin on.
This makes it possible to detect manifest updates automatically on the client-side.

coordinator/core/core.go
```diff
@@ -113,28 +120,34 @@ func NewCore(dnsNames []string, qv quote.Validator, qi quote.Issuer, sealer Seal
...
 	zapLogger.Info("loading state")
-	cert, privk, err := c.loadState()
+	rootCert, rootPrivK, intermediateCert, intermediatePrivK, err := c.loadState()
 	if err != nil {
 		if err != ErrEncryptionKey {
 			return nil, err
 		}
 		c.zaplogger.Error("Failed to decrypt sealed state. Processing with a new state. Use the /recover API endpoint to load an old state, or submit a new manifest to overwrite the old state. Look up the documentation for more information on how to proceed.")
-		cert, privk, err = c.generateCert(dnsNames)
+		c.mux.Lock()
```
thomasten, line 129: no need to lock the core in NewCore

coordinator/core/core.go
```diff
@@ -70,18 +72,23 @@ const (
...
+	Secrets             map[string]manifest.Secret
+	State               state
+	Activations         map[string]uint
 }
 
 // CoordinatorName is the name of the Coordinator. It is used as CN of the root certificate.
 const CoordinatorName string = "Marblerun Coordinator"
 
+// CoordinatorIntermediateName is the name of the Coordinator. It is used as CN of the intermediate certificate which is set when setting or updating a certificate.
+const CoordinatorIntermediateName string = "Marblerun Coordinator - Intermediate CA"
```
thomasten, line 90:
```suggestion
const coordinatorName string = "Marblerun Coordinator"

// CoordinatorIntermediateName is the name of the Coordinator. It is used as CN of the intermediate certificate which is set when setting or updating a certificate.
const coordinatorIntermediateName string = "Marblerun Coordinator - Intermediate CA"
```

coordinator/core/core.go
```diff
@@ -264,17 +293,14 @@ func (c *Core) sealState(recoveryData []byte) error {
 	return c.sealer.Seal(recoveryData, stateRaw)
 }
 
-func (c *Core) generateCert(dnsNames []string) (*x509.Certificate, *ecdsa.PrivateKey, error) {
-	defer c.mux.Unlock()
-	if err := c.requireState(stateUninitialized); err != nil {
-		return nil, nil, err
-	}
-
+func (c *Core) generateCert(dnsNames []string, parentCertificate *x509.Certificate, parentPrivateKey *ecdsa.PrivateKey) (*x509.Certificate, *ecdsa.PrivateKey, error) {
```
thomasten, line 296:
```suggestion
func generateCert(dnsNames []string, parentCertificate *x509.Certificate, parentPrivateKey *ecdsa.PrivateKey) (*x509.Certificate, *ecdsa.PrivateKey, error) {
```

coordinator/core/core.go
```diff
@@ -301,7 +327,15 @@ func (c *Core) generateCert(dnsNames []string) (*x509.Certificate, *ecdsa.Privat
...
 
-	certRaw, err := x509.CreateCertificate(rand.Reader, &template, &template, &privk.PublicKey, privk)
+	var certRaw []byte
+	if parentCertificate == nil || parentPrivateKey == nil {
+		certRaw, err = x509.CreateCertificate(rand.Reader, &template, &template, &privk.PublicKey, privk)
+	} else {
+		template.Issuer = parentCertificate.Issuer
+		template.Subject.CommonName = CoordinatorIntermediateName
+		certRaw, err = x509.CreateCertificate(rand.Reader, &template, parentCertificate, &privk.PublicKey, parentPrivateKey)
+	}
```
thomasten, line 337:
```suggestion
	if parentCertificate == nil {
		parentCertificate = &template
		parentPrivateKey = privk
	}
	certRaw, err := x509.CreateCertificate(rand.Reader, &template, parentCertificate, &privk.PublicKey, parentPrivateKey)
```
make behavior of this function more explicit by passing commonName as a parameter to generateCert
CreateCertificate does not use Issuer

coordinator/core/core.go
```diff
@@ -501,7 +535,7 @@ func (c *Core) generateCertificateForSecret(secret manifest.Secret, privKey cryp
 	}
 
 	template.IsCA = false
-	template.Issuer = c.cert.Subject
+	template.Issuer = parentCertificate.Subject
```
thomasten, line 538:
```suggestion
```

thomasten (changes requested):

thomasten:
> 
> 
> LGTM, we should think about returning the rootCert and intermediateCert on the client API.
> The client can decide which certificate to pin on.
> This makes it possible to detect manifest updates automatically on the client-side.

Yes, pls make this a new ticket.

Nirusu:
Ready for new review, now with an unit test added.

Regarding the new ticket: Should quoting over the whole chain be dealt with in this PR, or in a new one (with the API endpoints for getting the certificates)?

coordinator/core/clientapi.go
```diff
@@ -91,11 +91,26 @@ func (c *Core) SetManifest(ctx context.Context, rawManifest []byte) (map[string]
...
+	}
+
+	pemCertRoot := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: c.rootCert.Raw})
+	if len(pemCertRoot) <= 0 {
+		return "", nil, errors.New("pem.EncodeToMemory failed for root certificate")
+	}
+
+	// Include intermediate certificate if a manifest has been set
+	var pemCertIntermediate []byte
+	if c.intermediateCert != nil {
```
thomasten, line 106: This is not needed anymore, is it?

coordinator/core/clientapi.go
```diff
@@ -184,7 +221,17 @@ func (c *Core) UpdateManifest(ctx context.Context, rawUpdateManifest []byte) err
 
 	c.updateManifest = updateManifest
 	c.rawUpdateManifest = rawUpdateManifest
+	c.intermediateCert = intermediateCert
+	c.intermediatePrivK = intermediatePrivK
+
+	// Overwrite regenerated secrets in core
+	for name, secret := range regeneratedSecrets {
+		delete(c.secrets, name)
```
thomasten, line 229: seems superfluous to me

coordinator/core/clientapi.go
Nirusu: Yup, not needed anymore after the last commit. Will remove.

coordinator/core/clientapi.go
Nirusu: Actually I had to introduce it as a bug fix. Directly assigning the secret to the map in the core and thus, wanting to simply overwrite it, did not work on its own. 

Don't know why, but would be interesting to find out why it does not work. The most common solution you find online is using pointers in the map as value rather than a copy to the struct directly, but personally I would try to avoid using this solution as we do not really want to modify the secrets in the core often, just out of necessity on a manifest update.

thomasten: Works fine on my machine if I remove this line. Can you double-check and tell which test fails?

Nirusu: Double-checked, it actually does work. 

Guess this was an issue in the test then, as I did not deep-copy the map before comparing them initially, so I guess I was comparing pointers instead of the values in the structs, thus having this weird issue where delete fixed it. Later on I noticed this caused some other issues, so I deep-copied the map in the test which seems to fix all the issues I had.

So yeah, this can be removed.

Nirusu:
Okay, a new breakdown:

**Client Server**: Root CA 
**Marble Server** : Intermediate CA
**Marbles**: Intermediate CA + per-Marble Leaf Certificate
**Quote**: Root CA
**/quote API endpoint**: Returns the quote which was generated over the Root CA, but also returns the certificate chain consisting of the Intermediate CA + Root CA. `era` can pull this chain and extract the certificate an user may want to pin against. See: https://github.com/edgelesssys/era/pull/3
**Secrets (non-shared symmetric-key derivation)**: Derived from Root CA Private Key
**Secrets (certificates)**: Issued by Intermediate CA

m1ghtym0 (approved):
LGTM. Cert chain seems to be consistent with [RFC4346](https://tools.ietf.org/html/rfc4346#section-7.4.2). Good job!

thomasten (approved):
Nice! Pls clean up the commit history.

thomasten (approved):

m1ghtym0 (approved):

PR #91, opened 2021-02-03
AB#587 Add CLI rootca command

daniel-weisse:
Added CLI structure and rootca command

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -1,17 +1,8 @@
 package cmd
 
 import (
-	"fmt"
-	"io"
-	"net/http"
-	"os"
-
-	"github.com/edgelesssys/era/era"
 	"github.com/spf13/cobra"
 )
-
-var eraConfig string
-var insecureEra bool
 
 var rootCmd = &cobra.Command{
 	Use:   "marblerun",
@@ -26,38 +17,4 @@ func Execute() error {
 
 func init() {
 	rootCmd.AddCommand(newRootCACmd())
-}
-
-// verify the connection to the marblerun coordinator
-func verifyCoordinator(host string, configFilename string, insecure bool) (string, error) {
-	// skip verification if specified
-	if insecure {
-		fmt.Println("Warning: skipping quote verification")
-		return era.InsecureGetCertificate(host)
-	}
-
-	// get certificate using provided config
-	if configFilename != "" {
-		return era.GetCertificate(host, configFilename)
-	}
-
-	// get latest config from github if none specified
-	fmt.Println("No era config file specified, getting latest config from github.com/edgelesssys/marblerun/releases/latest/download/coordinator-era.json")
-	resp, err := http.Get("https://github.com/edgelesssys/marblerun/releases/latest/download/coordinator-era.json")
-	if err != nil {
-		return "", err
-	}
-	defer resp.Body.Close()
-	out, err := os.Create("era-config.json")
-	if err != nil {
-		return "", err
-	}
-	defer out.Close()
-	_, err = io.Copy(out, resp.Body)
-	if err != nil {
-		return "", err
-	}
-	fmt.Println("Got latest config")
-
-	return era.GetCertificate(host, "era-config.json")
 }
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -1,0 +1,82 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+
+	"github.com/edgelesssys/era/era"
+	"github.com/spf13/cobra"
+)
+
+var eraConfig string
+var insecureEra bool
+
+const eraString = `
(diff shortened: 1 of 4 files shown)


cli/cmd/rootCA.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"io/ioutil"
+
+	"github.com/spf13/cobra"
+)
+
+var rootCAFilename string
+
+func newRootCACmd() *cobra.Command {
+	cmd := &cobra.Command{
```
thomasten, line 13:
```suggestion
func newRootCACmd() *cobra.Command {
	var rootCAFilename string

	cmd := &cobra.Command{
```

cli/cmd/root.go
```diff
@@ -0,0 +1,82 @@
...
+	"io"
+	"net/http"
+	"os"
+
+	"github.com/edgelesssys/era/era"
+	"github.com/spf13/cobra"
+)
+
+var eraConfig string
+var insecureEra bool
```
thomasten, line 14: move those into newRootCACmd

cli/cmd/rootCA.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	cmd.Flags().StringVarP(&rootCAFilename, "output", "o", "marblerun.crt", "File to save the certificate to")
+	cmd.Flags().StringVar(&eraConfig, "era-config", "", eraString)
+	cmd.Flags().BoolVarP(&insecureEra, "insecure", "i", false, "Set to skip quote verification, needed when running in simulation mode")
+	return cmd
+}
+
+// CliRootCA gets the rootCA of the marblerun coordinator and saves it to
+func CliRootCA(host string, output string, configFilename string, insecure bool) error {
```
thomasten, line 32:
```suggestion
func cliRootCA(host string, output string, configFilename string, insecure bool) error {
```

cli/cmd/root.go
```diff
@@ -0,0 +1,82 @@
...
+	var cert string
+	var err error
+
+	if insecure {
+		// skip verification if specified
+		fmt.Println("Warning: skipping quote verification")
+		cert, err = era.InsecureGetCertificate(host)
+		if err != nil {
+			return "", err
+		}
```
thomasten, line 51:
```suggestion
		return era.InsecureGetCertificate(host)
```

cli/cmd/root.go
```diff
@@ -0,0 +1,82 @@
...
+
+		cert, err = era.GetCertificate(host, "era-config.json")
+		if err != nil {
+			return "", err
+		}
+	} else {
+		cert, err = era.GetCertificate(host, configFilename)
+		if err != nil {
+			return "", err
+		}
```
thomasten, line 79:
```suggestion
		return era.GetCertificate(host, configFilename)
```
Also move this block above the previous (with `if configFilename != ""`) so that you can reduce nesting, e.g.:
```go
if insecure {
	return ...
}
if configFilename != "" {
	return ...
}

// get latest ...

...

return era.Get...
```

cli/cmd/root.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/edgelesssys/era/era"
+	"github.com/spf13/cobra"
+)
+
+var eraConfig string
+var insecureEra bool
+
+const eraString = `
+Path to remote attestation config file in json format, if none provided the newest configuration will be loaded from github
+`
```
thomasten, line 18:
```suggestion
```
Any reason for not handling this usage string like the others?

thomasten (changes requested):

cli/cmd/root.go
```diff
@@ -0,0 +1,82 @@
...
+
+	rootCmd.AddCommand(newManifestCmd())
+	rootCmd.AddCommand(newNamespaceCmd())
+	rootCmd.AddCommand(newInstallCmd())
+	rootCmd.AddCommand(newRecoverCmd())
+	rootCmd.AddCommand(newRootCACmd())
+}
+
+// VerifyCoordinator verifies the connection using era
+func VerifyCoordinator(host string, configFilename string, insecure bool) (string, error) {
```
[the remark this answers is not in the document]
daniel-weisse: coordinator verification will be needed for other commands aswell, which is why the function is currently in root.go. Maybe this should be moved to its own file?

cli/cmd/root.go
daniel-weisse: same as verifyCoordinator, will be needed for other commands

cli/cmd/root.go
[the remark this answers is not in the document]
thomasten: Makes sense. Yes, we should probably move it to its own file because as I understand it, having root.go is some pattern of Cobra and we should then keep it free of our custom code.

m1ghtym0 (approved):
lgtm:-)

thomasten (approved):

PR #98, opened 2021-02-09
AB#609 Add CLI intermediate CA command

daniel-weisse:
This also creates a new parent command "certificate" which is used to then call either  "intermediate" or "root" (which was previously its own command "rootca") and changes "root" to only save the root certificate instead of both intermediate and root


cli/cmd/certificateIntermediate.go
```diff
@@ -0,0 +1,53 @@
...
+// cliCertificateIntermediate gets the intermediateCA of the Marblerun coordinator
+func cliCertificateIntermediate(host string, output string, configFilename string, insecure bool) error {
+	cert, err := verifyCoordinator(host, configFilename, insecure)
+	if err != nil {
+		return err
+	}
+
+	// Seperate rootCA from intermediateCA, if only one certificate save nothing and notify the user
+	var intermediateCA string
+	certSplit := strings.SplitAfter(cert, "-----END CERTIFICATE-----\n")
```
m1ghtym0, line 40: Please use the pem package for parsing here:
https://github.com/edgelesssys/era/blob/master/cmd/era/main.go#L65

cli/cmd/certificateRoot.go
```diff
@@ -0,0 +1,53 @@
...
+// cliCertificateRoot gets the rootCA of the Marblerun coordinator and saves it to a file
+func cliCertificateRoot(host string, output string, configFilename string, insecure bool) error {
+	cert, err := verifyCoordinator(host, configFilename, insecure)
+	if err != nil {
+		return err
+	}
+
+	// Seperate rootCA from intermediateCA, if only one certificate was returned save it as rootCA
+	var rootCA string
+	certSplit := strings.SplitAfter(cert, "-----END CERTIFICATE-----\n")
```
m1ghtym0, line 40: Same here.

m1ghtym0 (changes requested):
Please also add an option to output the whole chain.

cli/cmd/manifestUpdate.go
```diff
@@ -51,9 +51,15 @@ func cliManifestUpdate(manifestName string, host string, clCertFile string, clKe
...
-	if ok := certPool.AppendCertsFromPEM([]byte(caCert)); !ok {
+	if ok := certPool.AppendCertsFromPEM([]byte(caCert[0].Bytes)); !ok {
 		return errors.New("Failed to parse certificate")
 	}
+	// Add intermediate cert if applicable
+	if len(caCert) > 1 {
+		if ok := certPool.AppendCertsFromPEM([]byte(caCert[1].Bytes)); !ok {
+			return errors.New("Failed to parse certificate")
+		}
+	}
```
m1ghtym0, line 62: This is a bit confusing. caCert[0] is the intermediate, caCert[1] is the root cert (in case of a present intermediate).
Please change the comment accordingly.

cli/cmd/util.go
```diff
@@ -31,29 +32,35 @@ func verifyCoordinator(host string, configFilename string, insecure bool) (strin
...
-func restClient(cert string) (*http.Client, error) {
+func restClient(cert []*pem.Block) (*http.Client, error) {
 	// Set rootCA for connection to coordinator
 	certPool := x509.NewCertPool()
-	if ok := certPool.AppendCertsFromPEM([]byte(cert)); !ok {
-		return &http.Client{}, fmt.Errorf("failed to parse certificate")
+	if ok := certPool.AppendCertsFromPEM(pem.EncodeToMemory(cert[0])); !ok {
+		return &http.Client{}, fmt.Errorf("failed to parse root certificate")
+	}
+	// Add intermediate cert if applicable
```
m1ghtym0, line 59: Same here.

m1ghtym0 (changes requested):

m1ghtym0 (approved):

PR #101, opened 2021-02-15
kill child processes if integration test dies

thomasten:


m1ghtym0 (approved):

Nirusu (approved):
Seems to fix the problems I had when a test fails and the processes don't shut down correctly. Nice!

PR #102, opened 2021-02-16
AB#617: Use a consistent JSON output format for the API

Nirusu:
This... breaks stuff, so let's discuss that. Feedback is appreciated :)

## Idea
The general idea of this PR is to introduce a common JSON output format which always contains the status of the response, so that we do not receive an empty response after an user issued an API command such as setting a manifest and wondering what happened. Also, this might make error handling or integrating Marblerun into other tool sets more easier, as we always return a consistent JSON format, even for errors (unless the JSON marshaller fails badly...).

## Style
My idea was to follow the [JSend](https://github.com/omniti-labs/jsend) style in some way. There's not really a standard out there, but many APIs seem to follow a similar style. 

However, what the changes in this PR don't really use so far is the return type **fail**. Normally, we should return this if the server can handle the data, but the data is invalid. However, so far we really do not really differentiate error types, and even "internal" errors can result in returning an internal error with an HTTP 4xx code. Also, the **fail** type also "requires" the field `data`, but so far we would just embed one string inside which basically would work in the same way as **error** does... 

Therefore, so far, only **success** and **error** are implemented as types.

## Examples:
/status:
```json
{"status":"success","data":{"StatusCode":3,"StatusMessage":"Coordinator is running correctly and ready to accept marbles."}}
```
/quote:
```json
{"status":"success","data":{"Cert":"-----BEGIN CERTIFICATE-----\nMIIBwzCCAWmgAwIBAgIRAI2Wq/yoCXRqPu3NbbKwyB4wCgYIKoZIzj0EAwIwIDEe\nMBwGA1UEAxMVTWFyYmxlcnVuIENvb3JkaW5hdG9yMCAXDTIxMDIxNjA5NDUwN1oY\nDzIzMTMwNTI5MDkzMjI0WjAyMTAwLgYDVQQDEydNYXJibGVydW4gQ29vcmRpbmF0\nb3IgLSBJbnRlcm1lZGlhdGUgQ0EwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAAQy\nVl7JkbaLVPDO0qRx+kTpES+ltBFFkrj5kP9TWuOHpYz8l4SUta6ZQBh9/lOxHVKy\ne3jBJ/xmtkfIAJTrrJsno3AwbjAOBgNVHQ8BAf8EBAMCAoQwHQYDVR0lBBYwFAYI\nKwYBBQUHAwEGCCsGAQUFBwMCMA8GA1UdEwEB/wQFMAMBAf8wLAYDVR0RBCUwI4IJ\nbG9jYWxob3N0hwR/AAABhxAAAAAAAAAAAAAAAAAAAAABMAoGCCqGSM49BAMCA0gA\nMEUCIQC9VFtQsOyuj0FojOin5V8yOFlfIY93ixCXI6Ln2H16iAIgLIMjCddu11I0\nTVmKyLgahdAn55gksYEqWD2rD7npm1s=\n-----END CERTIFICATE-----\n-----BEGIN CERTIFICATE-----\nMIIBsTCCAVagAwIBAgIQZ7Ja3c7iTSUtGLob9KkQdjAKBggqhkjOPQQDAjAgMR4w\nHAYDVQQDExVNYXJibGVydW4gQ29vcmRpbmF0b3IwIBcNMjEwMjE2MDk0NTA3WhgP\nMjMxMzA1MjkwOTMyMjRaMCAxHjAcBgNVBAMTFU1hcmJsZXJ1biBDb29yZGluYXRv\ncjBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABAb//36qEfuWWYC98DlyFfDpy1Rv\nj+7mNjMjFrktcdZPLT6epUtNtvQN4blN0JCNnYuUAtMVeoQ7xJRFiqAeUMajcDBu\nMA4GA1UdDwEB/wQEAwIChDAdBgNVHSUEFjAUBggrBgEFBQcDAQYIKwYBBQUHAwIw\nDwYDVR0TAQH/BAUwAwEB/zAsBgNVHREEJTAjgglsb2NhbGhvc3SHBH8AAAGHEAAA\nAAAAAAAAAAAAAAAAAAEwCgYIKoZIzj0EAwIDSQAwRgIhANK15I85Th4l0NeMW7Qh\nlHEn80Tr3y420f08f0mLE7rIAiEAtD+jLNwy/2wGkQb5vK3l6ahypHXKW8mdB5hW\ny8/vEs8=\n-----END CERTIFICATE-----\n","Quote":""}}
```

/manifest (GET):
```json
{"status":"success","data":{"ManifestSignature":"3fff78e99dd9bd801e0a3a22b7f7a24a492302c4d00546d18c7f7ed6e26e95c3"}}
```

/manifest (POST, failed):
```json
{"status":"error","data":null,"message":"server is not in expected state"}
```

/manifest (POST, success, no secrets):
```json
{"status":"success","data":null}
```

/update (POST, failed):
```json
{"status":"error","data":null,"message":"unauthorized user"}
```

Think, you get the idea here. The old successful responses are embedded into "data", the error messages under "message".

## Issues
* It breaks quite some stuff, the documentation needs to be adjusted, and we need to adjust our tools here (e.g. `era`).
* Maybe we should add the type "fail"...?
* CLI stuff so far is untested, so maybe @daniel-weisse you want to take a look if the CLI stuff still works? Or maybe add some feedback on that in general, since you worked quite a bit recently with these API endpoints? :)

### DIFF
--- a/cli/cmd/manifestSet.go
+++ b/cli/cmd/manifestSet.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 
 	"github.com/spf13/cobra"
+	"github.com/tidwall/gjson"
 )
 
 func newManifestSet() *cobra.Command {
@@ -69,11 +70,18 @@ func cliManifestSet(manifestName string, host string, configFilename string, ins
 			return nil
 		}
 
+		response := gjson.GetBytes(respBody, "data")
+
+		// Skip outputting secrets if we do not get any recovery secrets back
+		if len(response.String()) == 0 {
+			return nil
+		}
+
 		// recovery secret was sent, print or save to file
 		if recover == "" {
-			fmt.Println(string(respBody))
+			fmt.Println(response.String())
 		} else {
-			if err := ioutil.WriteFile(recover, respBody, 0644); err != nil {
+			if err := ioutil.WriteFile(recover, []byte(response.String()), 0644); err != nil {
 				return err
 			}
 			fmt.Printf("Manifest successfully set, recovery data saved to: %s.\n", recover)
--- a/cli/cmd/manifestGet.go
+++ b/cli/cmd/manifestGet.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 
 	"github.com/spf13/cobra"
+	"github.com/tidwall/gjson"
 )
 
 func newManifestGet() *cobra.Command {
@@ -58,10 +59,11 @@ func cliManifestGet(targetFile string, host string, configFilename string, insec
 	switch resp.StatusCode {
 	case http.StatusOK:
 		respBody, err := ioutil.ReadAll(resp.Body)
+		manifestData := gjson.GetBytes(respBody, "data")
 		if err != nil {
 			return err
 		}
-		if err := ioutil.WriteFile(targetFile, respBody, 0644); err != nil {
+		if err := ioutil.WriteFile(targetFile, []byte(manifestData.String()), 0644); err != nil {
 			return err
 		}
 		fmt.Printf("Manifest written to: %s.\n", targetFile)
--- a/cli/cmd/recover.go
+++ b/cli/cmd/recover.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 
 	"github.com/spf13/cobra"
+	"github.com/tidwall/gjson"
 )
 
 type coordinatorResponse struct {
@@ -73,8 +74,9 @@ func cliRecover(host string, keyFile string, configFilename string, insecure boo
 		}
 
 		// if a response was sent another recovery key will be needed, print message to user
+		jsonResponse := gjson.GetBytes(respBody, "data")
 		var response coordinatorResponse
-		if err := json.Unmarshal(respBody, &response); err != nil {
+		if err := json.Unmarshal([]byte(jsonResponse.String()), &response); err != nil {
 			return err
 		}
 		fmt.Printf("%s \n", response)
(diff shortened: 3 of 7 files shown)


m1ghtym0:
LGTM:-)
The fail type would apply to scenarios like `/manifest` is called without providing a manifest in the request or the manifest does not match our specification?
Sounds useful to me. We can do this in a separate ticket and PR though.
I've added tickets for adjusting the cli, era, and the docs.
I can take care of the docs and assigned @3u13r for the cli, maybe @Nirusu can adjust era?

Nirusu:
Yep, I can take care of era. And then we'll with the `fail` type in another PR, sounds good to me.

cli/cmd/manifestSet.go
```diff
@@ -69,11 +70,12 @@ func cliManifestSet(manifestName string, host string, configFilename string, ins
 			return nil
 		}
 
+		response := gjson.GetBytes(respBody, "data")
```
daniel-weisse, line 73:
```suggestion
		response := gjson.GetBytes(respBody, "data")
		if response.String() == "" {
		    return nil
		}
```

daniel-weisse: Avoid printing empty line when no reovery secret was sent back

daniel-weisse (commented):
Avoid printing empty line when no reovery secret was sent back

Nirusu:
Should be fixed now.

m1ghtym0:
> I can take care of the docs and assigned @3u13r for the cli, maybe @Nirusu can adjust era?
Sorry, I meant to mention @daniel-weisse  here for the cli.

daniel-weisse (approved):

m1ghtym0 (approved):

PR #103, opened 2021-02-17
fix sample

thomasten:


Nirusu (approved):
Really wish we have automatic SGX-enabled testing at some point...

PR #107, opened 2021-02-22
Revised the samples for the Graphene support

b-64rry:
- Documentation (should be easier to grasp how to run apps)
- Pulled the nginx sample of Graphene
- Adapted the makefiles and the manifest template of the nginx sample to comply with Marblerun


samples/graphene-nginx/README.md
```diff
@@ -1,9 +1,13 @@
...
 
-*Prerequisite*: Graphene is set up and the nginx sample works.
+You can build the sample as follows:
+```sh
+export GRAPHENEDIR=[PATH To Your Graphene Folder]
+make
+```
 
-To marbleize the sample you must edit `nginx.manifest.template`. Please append or replace the following values:
+To marbleize the sample we edited `nginx.manifest.template`. Common things to change or add are:
```
m1ghtym0, line 10: Maybe elaborate a bit on what certain changes mean and why they are necessary.
From a user's perspective, I'm more curious about that when I want to port my own project and not can't or don't want to just copy-paste here.

samples/graphene-nginx/README.md
```diff
@@ -14,25 +18,32 @@ sgx.remote_attestation = 1
...
+
+We also replaced `sgx.trusted_files.cert` and `sgx.trusted_files.privkey` with
 ```toml
 sgx.protected_files.cert    = "file:install/conf/server.crt"
 sgx.protected_files.privkey = "file:install/conf/server.key"
 ```
-The server certificate will be injected by Marblerun, see [manifest.json](manifest.json).
+We delete these files before running nginx as they are handled by Marblerun.
+The server certificate, for instance, will be injected by Marblerun. See
+[manifest.json](manifest.json) on how this is specified.
```
m1ghtym0, line 33: Maybe also link to the secrets section in the Marblerun docs.

samples/graphene-nginx/README.md
```diff
@@ -14,25 +18,32 @@ sgx.remote_attestation = 1
...
 
-As you increased the `enclave_size`, you may need to decrease the number of `worker_processes` in `nginx-graphene.conf.template` to 1 or 2.
+As we increased the `enclave_size`, we might need to decrease the number of `worker_processes` in `nginx-graphene.conf.template` to 1 or 2.
```
m1ghtym0, line 35: Why is that?

m1ghtym0 (changes requested):
I took the time to review this from a user's perspective.
Something that was unclear to me with the copied files (nginx, ssl, Makefile): Are there any changes or is this just the copied example and we only patched the configs?
We should make this clear in the README or go a slightly different route:
Remove the copied files and provide a script/Makefile/... that pulls the original source and applies the provided patch.
We would keep the benefit of effortless running/testing of the sample, while not compromising on making it crystal clear what changes Marblerun requires.
Any thoughts on that @b-64rry @thomasten?
I've already mentioned it in the other comments, but we should elaborate a bit about the actual changes we do. As a user, I want to understand the reasoning behind them before I can comfortably port my own project. (Should add this to the docs anyway and might only link from here)

samples/graphene-nginx/README.md
b-64rry: I get your point on this. Thanks for the comments. I think this should go into the official  marblerun docs on how to "marblelize" your apps in general that run in Graphene or other mainstream LibOS implementations.

Then we can adjust this README, keep things small without losing 
concept and repeating ourselves. The sample is nice, not too complex and not too simple like a hello world and shows how we can achieve a synergy between our concepts and Graphene.

PR #116, opened 2021-03-17
AB#607 Add manual integration test for cli

daniel-weisse:
Files for the integration test are under test/cli

- Dockerfile: dockerfile to create the image used in the test, uses the hello-world sample
- hello-marble.yaml: k8s file to create a deployment and service for the helloworld image
- test-pod.yaml: k8s file to create a pod that does basically nothing, but can be used to check if correct values were injected
- test-manifest.json: manifest to be sent to the Marblerun coordinator during the test
- integration_test.sh: the test itself, probably not very pretty but does its job i hope

To run the test marblerun cli has to be installed and a minikube cluster has to be running
Start the test with `./integration_test.sh`

### DIFF
--- a/test/cli/integration_test.sh
+++ b/test/cli/integration_test.sh
@@ -1,10 +1,19 @@
 #!/bin/bash
 
 function wait_for_resource() {
+    grace=2
+    
     while true; do
         eval $1 > /dev/null
         if [ $? -eq 0 ]; then
             sleep 5
+            grace=2
+            continue
+        fi
+
+        if [ $grace -gt 0 ]; then
+            sleep 1
+            grace=$(($grace-1))
             continue
         fi
         break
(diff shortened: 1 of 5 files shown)


test/cli/integration_test.sh
```diff
@@ -0,0 +1,221 @@
...
+    exit 1
+fi
+
+kubectl delete deployments hello-marble
+kubectl delete svc hello-marble
+marblerun uninstall
+kubectl delete namespace marblerun
+wait_for_resource "kubectl get namespaces | grep marblerun"
+
+pkill kubectl
```
m1ghtym0, line 219:
```suggestion
kill %1 # kill kubectl port-forward
```
If you just want to kill the background port-forward, this might be more targeted.

m1ghtym0 (changes requested):

m1ghtym0 (approved):

PR #119, opened 2021-03-18
Add example and template for coordinator manifest

daniel-weisse:
The idea is to have a template users only need to fill in their data to create a valid manifest, and also provide an example of how a complete manifest would look like.

I am not sure if the current example-manifest.json provides good explanations (64 character string might better be 256 bytes hex format), or if the explanations are needed at all.
Maybe it is better to have a somewhat complete manifest explained in the docs that can be used in a demo application
For example we could provide a link to the manifest used by emojivoto (maybe even expand it to include RecoveryKeys and Admins fields to have a complete example).

### DIFF
--- a/example-manifest.json
+++ b/example-manifest.json
@@ -1,0 +1,95 @@
+{
+    "Packages": {
+        "<string, e.g. PackageName1>": {
+            "UniqueID": "<64 character string, not needed when SignerID is set>",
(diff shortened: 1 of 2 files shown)


m1ghtym0:
I like the idea of having an example and a template.
Maybe we should put them under `samples/manifest` and reference them in the README?

PR #120, opened 2021-03-22
AB#723 Add support for yml manifest and provide manifest templates

daniel-weisse:


m1ghtym0 (commented):
Please add a unit test for loading both file types

m1ghtym0 (commented):
Please add support to output in yaml as well.
We can use the `-o` parameter for this and output the result to stdout directly.

m1ghtym0 (commented):
I just noticed that we actually were missing something here.
The Client API `GET /manifest` actually does not return the whole manifest.
Only the manifest's signature is returned.
So it will get quite confusing if we output something called `manifest.json` either as json or yaml.
Instead we should only print the manifest signature to stdout.
Something like: `Manifest signature: 0x8080384028`
The CLI should output this signature when calling manifest set (it's just the sha256sum in the end).
We should also add a command `manifest signature` that prints the signature of a given yaml or json file

m1ghtym0 (approved):
lgtm:-)
@thomasten do you think it's ok that we calculate the signature locally?
imo it's just sha256 so we're not doing any magic here.

thomasten:
> @thomasten do you think it's ok that we calculate the signature locally?
> imo it's just sha256 so we're not doing any magic here.

I think it does not differ from the current workflow `sha256sum manifest.json`

PR #121, opened 2021-03-24
AB#677: add default env vars

katexochen:
This PR sets the default Marblerun ports to

- 2001 for gRPC calls to the Coordinator
- 4433 for http-REST client calls  to the Coordinator

These changes also apply to the Marblerun CLI, which previously used ports 25554 and 25555.

- [x] Ports might need to be changed in other repos and docs, like marblerun.sh, to those default ports.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -16,7 +16,9 @@ Marblerun is written entirely in Go and builds on Edgeless RT, which is written 
 
 ### Data Plane
 
-* [`marble`](marble): The data plane code written in Go
+* [`marble`](marble):
+	* [`config`](marble/config): Environment variables for configuration
+	* [`premain`](marble/config): The data plane code written in Go
 
 ## Build
 
@@ -36,10 +38,19 @@ make
 ### Run the Coordinator
 
 ```bash
-EDG_COORDINATOR_MESH_ADDR=localhost:2001 EDG_COORDINATOR_CLIENT_ADDR=localhost:4433 EDG_COORDINATOR_DNS_NAMES=localhost EDG_COORDINATOR_SEAL_DIR=$PWD OE_SIMULATION=1 erthost build/coordinator-enclave.signed
+OE_SIMULATION=1 erthost build/coordinator-enclave.signed
 ```
 
-*Note*: the Coordinator's state is sealed to `$PWD/sealed_data`. If you want a fresh restart remove this file first: `rm $PWD/sealed_data`.
+Per default, the Coordinator starts with the following default values. You can set your desired configuration by setting the environment variables.
+
+| Setting | Default Value | Environment Variable |
+| --- | --- | --- |
+| the listener address for the gRPC server | localhost:2001 |  EDG_COORDINATOR_MESH_ADDR |
+| the listener address for the HTTP server | localhost: 4433 | EDG_COORDINATOR_CLIENT_ADDR |
+| the DNS names for the cluster’s root certificate | localhost | EDG_COORDINATOR_DNS_NAMES |
+| the file path for storing sealed data | $PWD/marblerun-coordinator-data | EDG_COORDINATOR_SEAL_DIR |
+
+*Note*: The Coordinator's state is sealed to `$PWD/marblerun-coordinator-data/sealed_data`. If you want a fresh restart remove this file first: `rm $PWD/marblerun-coordinator-data/sealed_data`.
 
 ### Create a Manifest
 
@@ -107,7 +118,7 @@ Run a simple application.
     Run the server service in a new terminal:
 
     ```bash
-    EDG_MARBLE_COORDINATOR_ADDR=localhost:2001 EDG_MARBLE_TYPE=server EDG_MARBLE_UUID_FILE=$PWD/server_uuid EDG_MARBLE_DNS_NAMES=localhost EDG_TEST_ADDR=localhost:8001 OE_SIMULATION=1 erthost build/marble-test-enclave.signed
+    EDG_MARBLE_TYPE=server EDG_MARBLE_UUID_FILE=$PWD/server_uuid EDG_TEST_ADDR=localhost:8001 OE_SIMULATION=1 erthost build/marble-test-enclave.signed
     ```
 
     You should see: `[...] starting server`.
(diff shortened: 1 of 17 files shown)


m1ghtym0 (approved):

coordinator/config/config.go
```diff
@@ -7,20 +7,39 @@
...
 const DNSNames = "EDG_COORDINATOR_DNS_NAMES"
 
+// DNSNamesDefault are the default dns names for the coordinator's certificate
+const DNSNamesDefault = "localhost"
+
 // SealDir is the coordinator's file location to store the sealed state
 const SealDir = "EDG_COORDINATOR_SEAL_DIR"
 
+// SealDirDefault returns the coordinator's default file location to store the sealed state
+func SealDirDefault() string { return util.GetWd() + "/marblerun-coordinator-data" }
```
thomasten, line 39:
```suggestion
func SealDirDefault() string { return filepath.Join(util.GetWd(), "marblerun-coordinator-data") }
```

marble/config/config.go
```diff
@@ -7,14 +7,25 @@
...
 const DNSNames = "EDG_MARBLE_DNS_NAMES"
 
+// DNSNamesDefault returns the default alternative dns names for the marble's certificate
+func DNSNamesDefault() string { return util.MustGetEnv("EDG_MARBLE_TYPE") }
+
 // UUIDFile is the file path to store the marble's uuid
 const UUIDFile = "EDG_MARBLE_UUID_FILE"
+
+// UUIDFile is the default file path to store the marble's uuid
+func UUIDFileDefault() string { return util.GetWd() + "/uuid" }
```
thomasten, line 31:
```suggestion
func UUIDFileDefault() string { return filepath.Join(util.GetWd(), "uuid") }
```

marble/premain/premain_test.go
```diff
@@ -53,6 +53,7 @@ func TestPreMain(t *testing.T) {
 
 	issuer := quote.NewMockIssuer()
 
+	require.NoError(os.Setenv("EDG_CWD", "test"))
```
thomasten, line 56:
```suggestion
```

util/util.go
```diff
@@ -32,15 +32,23 @@ func DeriveKey(secret, salt []byte, length uint) ([]byte, error) {
...
 }
 
-// MustGetenv returns the environment variable `name` if it exists or panics otherwise
-func MustGetenv(name string) string {
-	value := os.Getenv(name)
-	if len(value) == 0 {
+// MustGetEnv returns the environment variable `name` if it exists or panics otherwise
+func MustGetEnv(name string) string {
+	value, found := os.LookupEnv(name)
+	if !found {
```
thomasten, line 38:
```suggestion
	value := os.Getenv(name)
	if len(value) == 0 {
```
failing on empty env var is on purpose

util/util.go
```diff
@@ -32,15 +32,23 @@ func DeriveKey(secret, salt []byte, length uint) ([]byte, error) {
 	return key, nil
 }
 
-// MustGetenv returns the environment variable `name` if it exists or panics otherwise
-func MustGetenv(name string) string {
-	value := os.Getenv(name)
-	if len(value) == 0 {
+// MustGetEnv returns the environment variable `name` if it exists or panics otherwise
+func MustGetEnv(name string) string {
```
thomasten, line 36:
```suggestion
// MustGetenv returns the environment variable `name` if it exists or panics otherwise
func MustGetenv(name string) string {
```
suggest naming the functions MustGetenv, Getenv, and MustGetwd in accordance with os package

util/util.go
```diff
@@ -81,3 +89,18 @@ func EncryptOAEP(pub *rsa.PublicKey, plaintext []byte) ([]byte, error) {
 func DecryptOAEP(priv *rsa.PrivateKey, ciphertext []byte) ([]byte, error) {
 	return rsa.DecryptOAEP(sha256.New(), rand.Reader, priv, ciphertext, nil)
 }
+
+// GetWd returns the current working directory and panics if it cannot be dcetermined.
+func GetWd() string {
+	// If marblerun runs in an enclave, EDG_CWD should be set.
+	wd, found := os.LookupEnv("EDG_CWD")
+	if found {
```
thomasten, line 97:
```suggestion
	if wd, ok := os.LookupEnv("EDG_CWD"); ok {
```

BUILD.md
```diff
@@ -36,10 +38,19 @@ make
...
+OE_SIMULATION=1 erthost build/coordinator-enclave.signed
 ```
 
-*Note*: the Coordinator's state is sealed to `$PWD/sealed_data`. If you want a fresh restart remove this file first: `rm $PWD/sealed_data`.
+Per default, the Coordinator starts with the following default values. You can set your desired configuration by setting the environment variables.
+
+| Setting | Default Value | Environment Variable |
+| --- | --- | --- |
+| the listener address for the gRPC server | localhost:2001 |  EDG_COORDINATOR_MESH_ADDR |
+| the listener address for the HTTP server | localhost: 4433 | EDG_COORDINATOR_CLIENT_ADDR |
```
thomasten, line 49: a bit out of scope because not introduced with this PR but: We should talk about the purpose of the servers rather than their technical implementation

BUILD.md
```diff
@@ -107,7 +118,7 @@ Run a simple application.
     Run the server service in a new terminal:
 
     ```bash
-    EDG_MARBLE_COORDINATOR_ADDR=localhost:2001 EDG_MARBLE_TYPE=server EDG_MARBLE_UUID_FILE=$PWD/server_uuid EDG_MARBLE_DNS_NAMES=localhost EDG_TEST_ADDR=localhost:8001 OE_SIMULATION=1 erthost build/marble-test-enclave.signed
+    EDG_MARBLE_TYPE=server EDG_MARBLE_UUID_FILE=$PWD/server_uuid EDG_TEST_ADDR=localhost:8001 OE_SIMULATION=1 erthost build/marble-test-enclave.signed
```
thomasten, line 121: we should also simplify the readmes of the samples

thomasten (changes requested):

marble/config/config.go
```diff
@@ -7,14 +7,29 @@
...
+const CoordinatorAddrDefault = "localhost:2001"
+
 // Type is the marble's type used for attestation with the coordinator
 const Type = "EDG_MARBLE_TYPE"
 
 // DNSNames are the alternative dns names for the marble's certificate
 const DNSNames = "EDG_MARBLE_DNS_NAMES"
 
+// DNSNamesDefault returns the default alternative dns names for the marble's certificate
+func DNSNamesDefault() string { return util.MustGetenv("EDG_MARBLE_TYPE") }
```
thomasten, line 29: Should we also include localhost here? E.g., to simplify the samples' Readmes. (also @m1ghtym0)

m1ghtym0: Sounds good to me

katexochen: I do not see how this would simplify the samples, since the EDG_MARBLE_TYPE does not have a default value and needs to be set anyway, there should not be a difference for the sample between using this type or "localhost". Or am I missing something?

thomasten: I think we need  to have localhost as one of the DNS names for the samples to work outside k8s, so the user would need to set both MARBLE_TYPE and DNS_NAMES.

thomasten (approved):

m1ghtym0 (approved):

PR #122, opened 2021-03-24
AB#729 Add version command

daniel-weisse:
Adds a version command to the CLI
Also allows the user to specify what version of Marblerun to install and fixes an issue where the Marblerun helm chart needed to be manually updated for the CLI to install the latest release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.11)
 
 project(coordinator)
+project(marblerun VERSION 0.3.0)
+string(TIMESTAMP date)
 find_package(OpenEnclave CONFIG REQUIRED)
 
 if (NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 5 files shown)


m1ghtym0 (commented):
I don't have experience with build-time version strings.
Would like to know @thomasten's opinion here.
Other than that lgtm:-)

thomasten:
Me neither. But I'd propose to define the version as `project(marblerun VERSION 0.3.0)` and then use CMAKE_PROJECT_VERSION (plus the timestamp if you like) to construct *one* version string and inject them in both cli and coordinator. (And let coordinator print it on start.)

CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.11)
 
 project(coordinator)
```
thomasten, line 3:
```suggestion
```

CMakeLists.txt
```diff
@@ -23,12 +25,14 @@ add_custom_command(
 add_custom_target(coordinatorlib
   ertgo build ${TRIMPATH} -buildmode=c-archive -tags enclave
   -o libcoordinator.a
+  -ldflags "-X 'main.Version=${PROJECT_VERSION}' -X 'main.BuildDate=${date}'"
```
thomasten, line 28: I informed myself about Project Oak at the weekend. They took some effort to achieve reproducible builds. I think this would be desirable for marblerun in the long term, too. Maybe we should drop build date. Maybe use git commit hash instead. What's your opinion? (Also @m1ghtym0)

thomasten (changes requested):

CMakeLists.txt
daniel-weisse: git commit hash sounds good, probably gives a better indication of when and what was built

thomasten (approved):

PR #123, opened 2021-03-25
Fix/misc

thomasten:
Fixed time calculation in TestActivate. I don't understand why the calculation in the tests leads to a different result than in the tested code (and didn't try too hard to figure it out), but it has to do with daylight saving time.
Other fixes are self-explanatory.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -142,7 +142,7 @@ func getRepo(name string, url string, settings *cli.EnvSettings) error {
 	repoFile := settings.RepositoryConfig
 
 	// Ensure the file directory exists as it is required for file locking
-	err := os.MkdirAll(filepath.Dir(repoFile), os.ModePerm)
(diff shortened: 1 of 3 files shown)


m1ghtym0 (approved):

cli/cmd/install.go
```diff
@@ -142,7 +142,7 @@ func getRepo(name string, url string, settings *cli.EnvSettings) error {
 	repoFile := settings.RepositoryConfig
 
 	// Ensure the file directory exists as it is required for file locking
-	err := os.MkdirAll(filepath.Dir(repoFile), os.ModePerm)
+	err := os.MkdirAll(filepath.Dir(repoFile), 0755)
```
Nirusu, line 145: Why 0755 here, but 0644 down below in line 194?

thomasten: Because this is a directory and you can't do much with a directory without X.

Nirusu (approved):

PR #124, opened 2021-03-25
Fix seg fault on marble activation

daniel-weisse:
There was an error where a segmentation fault occurred when a marble sent an activation request to the coordinator but the "Parameters" field in the Marblerun manifest was left empty. 
The coordinator would then try reading a value from a nil pointer and crash
This fixes that error by setting the Parameters field to default values if it was not provided by the Manifest (same as if the Manifest had specified '"Parameters": {}')

### DIFF
--- a/coordinator/manifest/manifest.go
+++ b/coordinator/manifest/manifest.go
@@ -62,7 +62,11 @@ func (m Manifest) Check(ctx context.Context, zaplogger *zap.Logger) error {
 	// if len(m.Infrastructures) <= 0 {
 	// 	return errors.New("no allowed infrastructures defined")
 	// }
-	for _, marble := range m.Marbles {
+	for idx, marble := range m.Marbles {
+		if marble.Parameters == nil {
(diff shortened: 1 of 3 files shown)


m1ghtym0 (approved):

coordinator/manifest/manifest.go
```diff
@@ -62,7 +62,11 @@ func (m Manifest) Check(ctx context.Context, zaplogger *zap.Logger) error {
 	// if len(m.Infrastructures) <= 0 {
 	// 	return errors.New("no allowed infrastructures defined")
 	// }
-	for _, marble := range m.Marbles {
+	for idx, marble := range m.Marbles {
+		if marble.Parameters == nil {
```
thomasten, line 66: Unit test is missing. In general, write a unit test and make sure it fails before fixing a bug. (See our workflow.)

thomasten (changes requested):

thomasten (approved):
thanks!

PR #126, opened 2021-04-01
AB#695: Automatically adapt Graphene manifest for Marblerun's premain

Nirusu:
So, this became quite the project.

This PR adds a feature to the Marblerun CLI in which Marblerun automatically tries to adapt an existing Graphene manifest to use Marblerun's premain & make the required changes.

For use with Graphene's examples, this seems to work mostly fine for the `preload` case with a minimal Marblerun manifest (basically only declaring the Package with MRENCLAVE and a Marble using this package), and also seems to work mostly fine for the `spawn` case when an adequate Marblerun manifest with correct argv provisioning was supplied to the coordinator.

Usage (`spawn`):
```bash
./marblerun graphene-prepare spawn <path>
```
Usage (`preload`):
```bash
./marblerun graphene-prepare preload <path>
```

Whereas <path> = path to the Graphene project manifest.template file

Screenshot of how it looks like:
Screenshot of graphene-prepare

Feedback appreciated :) Especially ideas to simplify the code, it's quite nested unfortunately...

### DIFF
--- a/cli/cmd/graphenePrepare.go
+++ b/cli/cmd/graphenePrepare.go
@@ -18,6 +18,15 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// premainNameSpawn is the name of the premain executable used for the 'spawn' method
+const premainNameSpawn = "premain-graphene"
+
+// premainNamePreload is the name of the premain shared library used for the 'preload' method
+const premainNamePreload = "premain-graphene.so"
+
+// uuidName is the file name of a Marble's uuid
+const uuidName = "uuid"
+
 type diff struct {
 	manifestEntry   string
 	alreadyExisting bool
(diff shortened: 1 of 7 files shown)


daniel-weisse (approved):
Looks good to me, can you squash the commits before merging?

thomasten (changes requested):
General structure seems fine, some minor simplification suggestions below. And some bugs, you may add tests for them before fixing.

cli/cmd/util.go
```diff
@@ -108,3 +112,21 @@ func getKubernetesInterface() (*kubernetes.Clientset, error) {
...
+	reader := bufio.NewReader(stdin)
+	response, err := reader.ReadString('\n')
+	if err != nil {
+		return false, err
+	}
+
+	response = strings.ToLower(strings.TrimSpace(response))
+
+	if response != "y" && response != "yes" {
+		fmt.Println("Aborting.")
```
thomasten, line 127: We shouldn't print this here. It limits situations in which this function can be used.

cli/cmd/graphenePrepare.go
```diff
@@ -0,0 +1,391 @@
...
+func appendAndReplace(changeDiffs []diff, manifestContent []byte) ([]byte, error) {
+	newManifestContent := manifestContent
+
+	var firstAdditionDone bool
+	for _, value := range changeDiffs {
+		if value.alreadyExists {
+			// If a value was previously existing, we replace the existing entry
+			key := strings.Split(value.manifestEntry, " =")
+			regexKey := strings.ReplaceAll(key[0], ".", "\\.")
+			regex := regexp.MustCompile("(?m)^" + regexKey + "\\s?=.*$")
```
thomasten, line 352:
```suggestion
			regex := regexp.MustCompile("(?m)^\\s*" + regexKey + "\\s*=.*$")
```
should we accept arbitrary whitespaces here?

Nirusu: I think it's not bad to support it. Might help when a manifest is slightly misformatted (douple spaces, lack of spaces etc.), but we're still sure that this is the right line to replace.

Nirusu: Might help for badly formatted files where we still know the exact line to replace. Any ideas where this could go wrong and cause mismatches?

Otherwise we could limit it to a certain number, though I think that would be an odd solution.

thomasten (approved):

cli/cmd/graphenePrepare.go
thomasten: Don't see mismatches, but am not confident. Decide yourself to change or keep as is.

PR #130, opened 2021-04-07
Miscellaneous fixes for CLI

daniel-weisse:
This PR fixes a few small things with the CLI:

* Parsing of intermediate certificate for the update command lead to an error
* Success message was printed twice when a recovery secret was returned and saved to file using `manifest set`
* `recover` was printing brackets when printing a returned status message

`recover` used to get its arguments in the following order:
* `marblerun recover <IP:PORT> <recovery_key>`

this was changed to be more in line with the rest of the CLI by passing IP:PORT as the second argument:
* `marblerun recover <recovery_key> <IP:PORT>`


m1ghtym0 (approved):
lgtm:-) Please also update the docs

PR #133, opened 2021-04-09
AB#730 Add cli check and precheck commands

daniel-weisse:
This PR adds two new commands to the CLI:

* `precheck`: checks if a cluster supports SGX
  * If no node of the cluster has SGX supports the command suggests installing in simulation mode
  * If a cluster supports SGX it will show how many nodes do so
  * Currently this checks only if the nodes provide these resources (which all have to be non zero for the node to be considered having SGX support) :
    * `sgx.intel.com/epc`
    * `sgx.intel.com/enclave`
    * `sgx.intel.com/provision`
  * If more implementations come up the code is structured in a way to easily add checks for them

* `check`: checks if the marblerun-coordinator and/or the marble-injector have been deployed on the cluster
  * If either the coordinator or the injector exists the command checks how many replicas the deployment is supposed to have and waits until all replicas have the `available` status

### DIFF
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -1,0 +1,99 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func newCheckCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "check",
+		Short: "Check the status of Marbleruns control plane",
+		Long:  `Check the status of Marbleruns control plane`,
+		Args:  cobra.NoArgs,
+		RunE: func(cobracmd *cobra.Command, args []string) error {
+			kubeClient, err := getKubernetesInterface()
+			if err != nil {
+				return err
+			}
+			return cliCheck(kubeClient)
+		},
+		SilenceUsage: true,
+	}
+
+	return cmd
+}
+
+// check if marblerun control-plane deployments are ready to use
+func cliCheck(kubeClient kubernetes.Interface) error {
+	err := checkDeploymentStatus(kubeClient, "marble-injector", "marblerun")
+	if err != nil {
+		return err
+	}
+
+	err = checkDeploymentStatus(kubeClient, "marblerun-coordinator", "marblerun")
+	if err != nil {
+		return err
+	}
+
+	// Add checks for other control plane deployments here
+	//
+	// err = checkDeploymentStatus(kubeClient, "some-control-plane-component", "marblerun")
+	//
+
+	return nil
+}
+
+// checkDeploymentStatus checks if a deployment is installed on the cluster
+// If it is, this function will wait until all replicas have the "available" status (ready for at least minReadySeconds)
+// Current status is continuously printed on screen
+func checkDeploymentStatus(kubeClient kubernetes.Interface, deploymentName string, namespace string) error {
+	_, err := kubeClient.AppsV1().Deployments(namespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	if errors.IsNotFound(err) {
+		fmt.Printf("%s is not installed on this cluster\n", deploymentName)
+	} else {
+		var podsReady string
+		deploymentReady := false
+		for !deploymentReady {
+			deploymentReady, podsReady, err = deploymentIsReady(kubeClient, deploymentName, namespace)
+			if err != nil {
+				return err
+			}
+
+			updateString := fmt.Sprintf("%s pods ready: %s", deploymentName, podsReady)
+			for i := 0; i < len(updateString); i++ {
+				fmt.Printf(" ")
+			}
+
+			fmt.Printf("\r%s", updateString)
+			time.Sleep(1)
+		}
+		fmt.Println()
+	}
+
+	return nil
+}
+
+// deploymentIsReady checks on the status of a single deployment and returns the number of available pods in the form "available/total"
+func deploymentIsReady(kubeClient kubernetes.Interface, deploymentName string, namespace string) (bool, string, error) {
+	deployment, err := kubeClient.AppsV1().Deployments(namespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
+	if err != nil {
+		return false, "", err
+	}
+
+	status := fmt.Sprintf("%d/%d", deployment.Status.AvailableReplicas, deployment.Status.Replicas)
+	if deployment.Status.Replicas == deployment.Status.AvailableReplicas {
+		return true, status, nil
+	}
+
+	return false, status, nil
+}
--- a/cli/cmd/check.go
+++ b/cli/cmd/check.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -12,6 +13,8 @@ import (
 )
 
 func newCheckCmd() *cobra.Command {
+	var timeout uint
+
 	cmd := &cobra.Command{
 		Use:   "check",
 		Short: "Check the status of Marbleruns control plane",
@@ -22,30 +25,24 @@ func newCheckCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			return cliCheck(kubeClient)
+			return cliCheck(kubeClient, timeout)
 		},
 		SilenceUsage: true,
 	}
 
+	cmd.Flags().UintVar(&timeout, "timeout", 60, "Time to wait before aborting in seconds")
 	return cmd
 }
 
-// check if marblerun control-plane deployments are ready to use
-func cliCheck(kubeClient kubernetes.Interface) error {
-	err := checkDeploymentStatus(kubeClient, "marble-injector", "marblerun")
-	if err != nil {
+// cliCheck if marblerun control-plane deployments are ready to use
+func cliCheck(kubeClient kubernetes.Interface, timeout uint) error {
+	if err := checkDeploymentStatus(kubeClient, "marble-injector", "marblerun", timeout); err != nil {
 		return err
 	}
 
-	err = checkDeploymentStatus(kubeClient, "marblerun-coordinator", "marblerun")
-	if err != nil {
+	if err := checkDeploymentStatus(kubeClient, "marblerun-coordinator", "marblerun", timeout); err != nil {
 		return err
 	}
-
-	// Add checks for other control plane deployments here
-	//
-	// err = checkDeploymentStatus(kubeClient, "some-control-plane-component", "marblerun")
-	//
 
 	return nil
 }
@@ -53,7 +50,7 @@ func cliCheck(kubeClient kubernetes.Interface) error {
 // checkDeploymentStatus checks if a deployment is installed on the cluster
 // If it is, this function will wait until all replicas have the "available" status (ready for at least minReadySeconds)
 // Current status is continuously printed on screen
-func checkDeploymentStatus(kubeClient kubernetes.Interface, deploymentName string, namespace string) error {
+func checkDeploymentStatus(kubeClient kubernetes.Interface, deploymentName string, namespace string, timeout uint) error {
 	_, err := kubeClient.AppsV1().Deployments(namespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
 	if err != nil && !errors.IsNotFound(err) {
 		return err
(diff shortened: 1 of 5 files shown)


cli/cmd/precheck.go
```diff
@@ -0,0 +1,93 @@
...
+
+	numNodes := len(nodes.Items)
+	supportedNodes := 0
+
+	// Iterate over all nodes in the cluster and check their SGX support
+	for i := 0; i < numNodes; i++ {
+		if nodeSupportsSGX(nodes.Items[i].Status.Capacity) {
+			supportedNodes++
+		}
+	}
```
m1ghtym0, line 52:
```suggestion
	for _, node := range nodes.Items {
		if nodeSupportsSGX(node.Status.Capacity) {
			supportedNodes++
		}
	}
```

cli/cmd/precheck.go
```diff
@@ -0,0 +1,93 @@
...
+	return cmd
+}
+
+func cliCheckSGXSupport(kubeClient kubernetes.Interface) error {
+	nodes, err := kubeClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	numNodes := len(nodes.Items)
```
m1ghtym0, line 44:
```suggestion
```

cli/cmd/precheck.go
```diff
@@ -0,0 +1,93 @@
...
+// nodeSupportsSGX checks if a single cluster node supports SGX in some way
+// Checks for different implementations of kubernetes SGX should be put here (e.g. different resource definitions than the one used by Azure)
+func nodeSupportsSGX(capacityInfo corev1.ResourceList) bool {
+	if nodeSupportsAzureSGX(capacityInfo) {
+		return true
+	}
+
+	// if nodeSupports[SomeCloudProvider]SGX(capacityInfo) {
+	// 	return true
+	// }
```
m1ghtym0, line 79:
```suggestion
```

Similar to my comment above, I believe one can infer this from the existing case.

cli/cmd/precheck.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		fmt.Printf("Cluster supports SGX on %d %s\n", supportedNodes, nodeString)
+		fmt.Println("To install Marblerun run [marblerun install]")
+	}
+
+	return nil
+}
+
+// nodeSupportsSGX checks if a single cluster node supports SGX in some way
+// Checks for different implementations of kubernetes SGX should be put here (e.g. different resource definitions than the one used by Azure)
```
m1ghtym0, line 71:
```suggestion
// Checks for different implementations of SGX device plugins should be put here (e.g. different resource definitions than the one used by Azure)
```

Has Azure switched to the Intel plugin? If so we should name it accordingly.

cli/cmd/precheck_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	err = cliCheckSGXSupport(testClient)
+	require.NoError(err)
+
+	testNodeSGX.ObjectMeta.Name = "sgx-node-2"
+	_, err = testClient.CoreV1().Nodes().Create(context.TODO(), testNodeSGX, metav1.CreateOptions{})
+	require.NoError(err)
+
+	err = cliCheckSGXSupport(testClient)
+	require.NoError(err)
```
m1ghtym0, line 103:
```suggestion
	assert.NoError(err)
```

cli/cmd/precheck_test.go
```diff
@@ -0,0 +1,104 @@
...
+				intelEpc:       resource.MustParse("500"),
+				intelProvision: resource.MustParse("10"),
+			},
+		},
+	}
+	_, err = testClient.CoreV1().Nodes().Create(context.TODO(), testNodeSGX, metav1.CreateOptions{})
+	require.NoError(err)
+
+	err = cliCheckSGXSupport(testClient)
+	require.NoError(err)
```
m1ghtym0, line 96:
```suggestion
	assert.NoError(err)
```

cli/cmd/precheck_test.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+	}
+	_, err := testClient.CoreV1().Nodes().Create(context.TODO(), testNode, metav1.CreateOptions{})
+	require.NoError(err)
+
+	_, err = testClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	require.NoError(err)
+
+	err = cliCheckSGXSupport(testClient)
+	require.NoError(err)
```
m1ghtym0, line 77:
```suggestion
	assert.NoError(err)
```

cli/cmd/precheck_test.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+	}
+	_, err = testClient.CoreV1().Nodes().Create(context.TODO(), testNodeSGX, metav1.CreateOptions{})
+	require.NoError(err)
+
+	nodes, err = testClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	require.NoError(err)
+
+	supportsSGX = nodeSupportsSGX(nodes.Items[0].Status.Capacity)
+	assert.True(supportsSGX, "Function returned false for nodes supporting SGX")
```
m1ghtym0, line 58:
```suggestion
	assert.True(supportsSGX, "function returned false for nodes supporting SGX")
```

cli/cmd/precheck_test.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+	}
+	_, err := testClient.CoreV1().Nodes().Create(context.TODO(), testNode, metav1.CreateOptions{})
+	require.NoError(err)
+
+	nodes, err := testClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	require.NoError(err)
+
+	supportsSGX := nodeSupportsSGX(nodes.Items[0].Status.Capacity)
+	assert.False(supportsSGX, "Function returned true for nodes not supporting SGX")
```
m1ghtym0, line 33:
```suggestion
	assert.False(supportsSGX, "function returned true for nodes not supporting SGX")
```

cli/cmd/check_test.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+	}
+
+	_, err = testClient.AppsV1().Deployments(namespace).Create(context.TODO(), testDeployment, metav1.CreateOptions{})
+	require.NoError(err)
+
+	ready, status, err := deploymentIsReady(testClient, deploymentName, namespace)
+	require.NoError(err)
+	assert.False(ready, "function returned true when deployment was not ready")
+	assert.Equal("0/1", status, fmt.Sprintf("Expected 0/1 ready pods but got %s", status))
```
m1ghtym0, line 44:
```suggestion
	assert.Equal("0/1", status, fmt.Sprintf("expected 0/1 ready pods but got %s", status))
```

cli/cmd/check_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	testDeployment.Status.UnavailableReplicas = 0
+	testDeployment.Status.AvailableReplicas = 1
+	_, err = testClient.AppsV1().Deployments(namespace).UpdateStatus(context.TODO(), testDeployment, metav1.UpdateOptions{})
+	require.NoError(err)
+
+	ready, status, err = deploymentIsReady(testClient, deploymentName, namespace)
+	require.NoError(err)
+	assert.True(ready, "function returned false when deployment was ready")
+	assert.Equal("1/1", status, fmt.Sprintf("Expected 1/1 ready pods but got %s", status))
```
m1ghtym0, line 54:
```suggestion
	assert.Equal("1/1", status, fmt.Sprintf("expected 1/1 ready pods but got %s", status))
```

cli/cmd/precheck.go
```diff
@@ -0,0 +1,93 @@
...
+func nodeSupportsSGX(capacityInfo corev1.ResourceList) bool {
+	if nodeSupportsAzureSGX(capacityInfo) {
```
m1ghtym0, line 73:
```suggestion
	if nodeHasAzureDevPlugin(capacityInfo) {
```

cli/cmd/precheck.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// if nodeSupports[SomeCloudProvider]SGX(capacityInfo) {
+	// 	return true
+	// }
+
+	return false
+}
+
+// nodeSupportsAzureSGX checks if nodes in the cluster contain Azures SGX definitions
+func nodeSupportsAzureSGX(capacityInfo corev1.ResourceList) bool {
```
m1ghtym0, line 85:
```suggestion
// nodeHasAzureDevPlugin checks if a node has the Azures SGX device plugin installed
func nodeHasAzureDevPlugin(capacityInfo corev1.ResourceList) bool {
```

cli/cmd/precheck_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestNodeSupportsAzureSGX(t *testing.T) {
```
m1ghtym0, line 15:
```suggestion
func TestNodeHasAzureDevPlugin(t *testing.T) {
```

m1ghtym0 (changes requested):

cli/cmd/check.go
```diff
@@ -0,0 +1,99 @@
...
+		for !deploymentReady {
+			deploymentReady, podsReady, err = deploymentIsReady(kubeClient, deploymentName, namespace)
+			if err != nil {
+				return err
+			}
+
+			updateString := fmt.Sprintf("%s pods ready: %s", deploymentName, podsReady)
+			for i := 0; i < len(updateString); i++ {
+				fmt.Printf(" ")
+			}
```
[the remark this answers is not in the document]
daniel-weisse: If we run this command in a terminal the outpout is continuesly updated on a single line. The spaces are used to clear the current line. I have looked into it again and there seems to be a better way using escape codes
```golang
// save initial cursor position
fmt.Print("\033[s")

// print something
fmt.Printf("Text to clear")

// restore cursor position and clear line
fmt.Print("\033[u\033[K")
```

cli/cmd/precheck.go
daniel-weisse: Currently when creating a SGX kubernetes cluster on azure the Intel resource definitions are used. I will add a check for the previous `kubernetes.azure.com/sgx_epc_mem_in_MiB` plugin just in case this changes again / can be changed by users

cli/cmd/check.go
```diff
@@ -22,64 +25,79 @@ func newCheckCmd() *cobra.Command {
...
 	return nil
 }
 
 // checkDeploymentStatus checks if a deployment is installed on the cluster
 // If it is, this function will wait until all replicas have the "available" status (ready for at least minReadySeconds)
 // Current status is continuously printed on screen
-func checkDeploymentStatus(kubeClient kubernetes.Interface, deploymentName string, namespace string) error {
+func checkDeploymentStatus(kubeClient kubernetes.Interface, deploymentName string, namespace string, timeout uint) error {
+	var tries uint
+	var podsReady string
```
m1ghtym0, line 55:
```suggestion
```

cli/cmd/check.go
```diff
@@ -22,64 +25,79 @@ func newCheckCmd() *cobra.Command {
...
 	} else {
-		var podsReady string
 		deploymentReady := false
-		for !deploymentReady {
+
+		if isTTY {
+			// save current cursor position
+			fmt.Print("\033[s")
+		}
+
```
m1ghtym0, line 76:
```suggestion
	tries := 0
	deploymentReady := false
```

cli/cmd/check.go
```diff
@@ -22,64 +25,79 @@ func newCheckCmd() *cobra.Command {
...
 	_, err := kubeClient.AppsV1().Deployments(namespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
 	if err != nil && !errors.IsNotFound(err) {
 		return err
 	}
 	if errors.IsNotFound(err) {
 		fmt.Printf("%s is not installed on this cluster\n", deploymentName)
 	} else {
-		var podsReady string
 		deploymentReady := false
```
m1ghtym0, line 70:
```suggestion
```

cli/cmd/check.go
```diff
@@ -22,64 +25,79 @@ func newCheckCmd() *cobra.Command {
...
-		for !deploymentReady {
+
+		if isTTY {
+			// save current cursor position
+			fmt.Print("\033[s")
+		}
+
+		for !deploymentReady && tries < timeout {
 			deploymentReady, podsReady, err = deploymentIsReady(kubeClient, deploymentName, namespace)
```
m1ghtym0, line 78:
```suggestion
			deploymentReady, podsReady, err := deploymentIsReady(kubeClient, deploymentName, namespace)
```

m1ghtym0 (changes requested):

cli/cmd/check.go
daniel-weisse: `tries` needs to be declared as `uint` since its compared against `timeout` which is also of type `uint` to prevent users from setting negativ timeout durations.
`uint` defaults to `0`, so we don't need to explicitly assign this value before using `tries`.

cli/cmd/check.go
daniel-weisse: We need to declare `deploymentReady` since it is used to check if we should continue the following loop.
This also means we can not use `deploymentReady, podsReady, err := deploymentIsReady(...)` since this would declare `deploymentReady` at loop scope resulting in execution until the `timeout` limit is reached.

m1ghtym0: Oh something got mixed up in my review, I meant to remove podsReady because it's only used inside the loop.
That's why I also changed "=" to ":="

cli/cmd/precheck.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// nodeHasIntelDevPlugin checks if a node has the Intel SGX device plugin installed
+func nodeHasIntelDevPlugin(capacityInfo corev1.ResourceList) bool {
+	epcQuant := capacityInfo[intelEpc]
+	enclaveQuant := capacityInfo[intelEnclave]
+	provisionQuant := capacityInfo[intelProvision]
+	if epcQuant.Value() == 0 || enclaveQuant.Value() == 0 || provisionQuant.Value() == 0 {
+		return false
+	}
```
m1ghtym0, line 99:
```suggestion
	epcQuant, ok := capacityInfo[intelEpc]
	if !ok || epcQuant.Value() == 0 {
		return false
	}
	enclaveQuant, ok := capacityInfo[intelEnclave]
	if !ok || enclaveQuant.Value() == 0 {
		return false
	}
	provisionQuant, ok := capacityInfo[intelProvision]
	if !ok || provisionQuant.Value() == 0 {
		return false
	}
```
Not sure if `corev1.ResourceList` has a default value of 0 for non-existing ResouceNames?

m1ghtym0 (changes requested):

m1ghtym0 (approved):

PR #128, opened 2021-04-06
AB#757 Load correct era-config depending on coordinator version

daniel-weisse:
Instead of always loading the latest era-config, we now check what version the coordinator is running and try to download a matching config

PR #135, opened 2021-04-13
Check http status before saving era config to file

daniel-weisse:
As requested here: https://github.com/edgelesssys/marblerun/pull/128#discussion_r611415094

This PR adds a check to `verifyCoordinator` in `cli/cmd/util.go` which makes sure we received an `OK` status when downloading the specified era-config.

Previously any problems during the http get request which would not result in an error from the function (e.g. `404`) were not caught  and would result in errors later on. This should be fixed now.


m1ghtym0 (approved):

Nirusu (approved):
Thank you!

PR #136, opened 2021-04-14
Update to work correctly with new cli version

daniel-weisse:
This PR updates `test/cli/integration_test.sh` to work correctly with cli version 0.3.1.
Also adds a function to cleanup the cluster if a command fails

### DIFF
--- a/test/cli/integration_test.sh
+++ b/test/cli/integration_test.sh
@@ -20,18 +20,10 @@ function wait_for_resource() {
     done
 }
 
-function cleanup_error() {
+function cleanup_cluster() {
     marblerun uninstall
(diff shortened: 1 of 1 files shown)


test/cli/integration_test.sh
```diff
@@ -20,6 +20,20 @@ function wait_for_resource() {
     done
 }
 
+function cleanup_error() {
```
m1ghtym0, line 23: Can we simplify this with smth. like: https://unix.stackexchange.com/questions/124127/kill-all-descendant-processes ?

m1ghtym0 (approved):

m1ghtym0 (approved):

PR #138, opened 2021-04-15
AB#804 Add CLI complete command

daniel-weisse:
This PR adds a new command `completion`.
This new command generates a shell script which enables autocompletion for the cli.

I tested this with `bash` and `zsh` without encountering errors.
`fish` completion works but i received some errors when using the completion. If anyone uses the `fish` shell and would like to test it i would be glad to hear your feedback. If not i will remove support for `fish` before merging.

### DIFF
--- a/cli/cmd/completion.go
+++ b/cli/cmd/completion.go
@@ -1,0 +1,68 @@
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func newCompletionCmd() *cobra.Command {
+	example := `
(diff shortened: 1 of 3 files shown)


daniel-weisse:
There are apparently some problems with `fish` autocompletion in cobra at the moment and there is a PR to fix them: https://github.com/spf13/cobra/pull/1249
However, since i dont believe `fish` is a commonly used shell, it is easiest to drop support for now and maybe re-add it once the problems are fixed.

m1ghtym0 (approved):
LGTM:-)
Please add docs as well

PR #139, opened 2021-04-19
AB#782 refactor injector and cli install to be device plugin independant

daniel-weisse:
This PR changes the marble-injector to allow setting which resource-key to inject into starting pods.
Previously this was fixed to a preset value: `kubernetes.azure.com/sgx_epc_mem_in_MiB`
This made the injector inflexible and required a specific device plugin to be installed on a cluster. Now this value can be set on startup of the injector using the `sgxResource` flag.

The cli install command was also changed to reflect this new feature:
* The command will try to determine if the cluster supports SGX, the same way `precheck` does, and pass this value to the start options for the injector
  * If no support can be determined a default value will be used, this can be set using the `resource-key` flag
* This key will also be used to set tolerations and resource limits for the coordinator


cli/cmd/install.go
```diff
@@ -303,5 +334,45 @@ func verifyNamespace(namespace string, kubeClient kubernetes.Interface) error {
 	return nil
 }
 
+// getResourceKey checks what device plugin is providing SGX on the cluster and returns the corresponding resource key
+func getResourceKey(defaultKey string) (string, error) {
```
m1ghtym0, line 338:
```suggestion
func getSGXResourceKey(defaultKey string) (string, error) {
```

cli/cmd/install.go
```diff
@@ -23,117 +23,148 @@ import (
...
+	if !options.simulation {
+		// Since we want to potentially change a key in the "limits" map we have to remove keys that are different from the one we want to insert
+		// If we just set a new key the resulting deployment will potentially have 2 resource limits for different SGX definitions resulting in a pod that will never start
+		if presetLimits, ok := chart.Values["coordinator"].(map[string]interface{})["resources"].(map[string]interface{})["limits"].(map[string]interface{}); ok {
+			finalValues["coordinator"].(map[string]interface{})["resources"] = map[string]interface{}{
+				"limits": map[string]interface{}{},
+			}
+			for oldResourceKey := range presetLimits {
+				// Make sure the key we delete is not a common kubernetes resource type
+				if oldResourceKey != resourceKey && !(strings.Contains("cpu memory", oldResourceKey) || strings.Contains(oldResourceKey, "hugepages-")) {
```
m1ghtym0, line 147: This is dangerous, imagine someone using other custom resources.
The scenario you're trying to prevent is someone setting a `kubernetes.azure.com/sgx_epc_mem_in_MiB` limit in their k8s file but installing marblerun with the default resourceKey?
I'd suggest we disable the injection of limits and tolerations by default and if someone enables them, we leave it to them to set the correct resource key.

m1ghtym0 (changes requested):

cli/cmd/install.go
daniel-weisse: Do we instead just install the coordinator using the values provided by the helm chart?
Another approach could be checking the resource keys of the helm chart against a list of known sgx keys (azure, intel etc.) and then removing those, provided they are different from what we want to set as the new value. This should prevent accidentally removing custom resource requests a user might have set.

The change i would make is:
* check if the `resource-key` flag was set by a user or left empty
* if it is empty, check the cluster for sgx resources and return the key (`sgx.intel.com/epc`), if we cant find any return a default value
* if it was set, return that value as the key
* set the returned value as toleration for the coordinator
* if the coordinator has any sgx resource requests different from the returned key remove them

Then again we could just start the coorindator likei it is defined in the helm chart and only if a user sets the `resource-key` flag we change tolerations/resource limits

cli/cmd/csr_legay.go
```diff
@@ -70,11 +70,12 @@ func (crt *certificateLegacy) get() ([]byte, error) {
...
 // setCaBundle sets the CABundle field to the self signed rootCA generated by the handler
-func (crt *certificateLegacy) setCaBundle(values map[string]interface{}) error {
+func (crt *certificateLegacy) setCaBundle(values map[string]interface{}, resourceKey string) error {
 	caCertBytes := pem.EncodeToMemory(crt.caCert)
 	values["marbleInjector"] = map[string]interface{}{
-		"start":    true,
-		"CABundle": base64.StdEncoding.EncodeToString(caCertBytes),
+		"start":       true,
+		"CABundle":    base64.StdEncoding.EncodeToString(caCertBytes),
+		"resourceKey": resourceKey,
```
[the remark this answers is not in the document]
daniel-weisse: the idea was to have all injector changes to the helm chart in one place.
It probably makes more sense though to have this outside the setCABundle function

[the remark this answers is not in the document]
m1ghtym0: Yes, either refactor this function so it represents the "changes all helm charts things" or have different "setters".

cli/cmd/install.go
m1ghtym0: The downside of your suggestion would be that we need to update this code as well if we want to add a new plugin?
I don't believe this will happen very often with the official Intel plugin, so your suggestion would be fine with me:-)

m1ghtym0 (approved):
Did you run the integration test?

daniel-weisse:
Integration test finished without errors.
I did some manual tests for preset tolerations and resources, the current version seems to work fine without overwriting values that shouldnt be overwritten

m1ghtym0 (approved):

PR #141, opened 2021-04-21
Feat/improve samples

katexochen:


samples/helloc++/README.md
```diff
@@ -1,5 +1,5 @@
-# C++ sample
-This sample shows how to build a confidential C++ application and run it in Marblerun. This should serve as a blueprint for making existing applications Marblerun-ready or creating new ones.
+# How to create a C++ Marble
+This sample shows how to build a confidential C++ application and run it in Marblerun. This can serve you as a blueprint for making existing applications Marblerun-ready or creating new creating new [Marbles](https://www.marblerun.sh/docs/getting-started/marbles/). If you haven't already, [setup Marblerun](../../BUILD.md#build) to get ready.
```
m1ghtym0, line 2:
```suggestion
This sample shows how to build a confidential C++ application and run it in Marblerun. This can serve you as a blueprint for making existing applications Marblerun-ready or creating new [Marbles](https://www.marblerun.sh/docs/getting-started/marbles/). If you haven't already, [setup Marblerun](../../BUILD.md#build) to get ready.
```

m1ghtym0 (changes requested):

samples/graphene-hello/preload/README.md
```diff
@@ -19,31 +19,24 @@ Then get `mr_enclave` from the build output and set it as `UniqueID` in `manifes
...
 ## Run
-We assume that the Coordinator is run with the following environment variables:
-
-- EDG_COORDINATOR_MESH_ADDR=localhost:2001
-- EDG_COORDINATOR_CLIENT_ADDR=localhost:4433
-- EDG_COORDINATOR_DNS_NAMES=localhost
-- EDG_COORDINATOR_SEAL_DIR=$PWD
+Next, use the `erthost` command to start the Coordinator in a local simulated enclave:
+```sh
+OE_SIMULATION=1 erthost ../../../build/coordinator-enclave.signed
```
Nirusu, line 24: Why OE_SIMULATION=1? 

Graphene specifically requires SGX-enabled hardware. You cannot run Graphene applications in simulation mode. And if you already run the application on an SGX-enabled system (which I think we should assume for the samples), why specifically run the Cooridinator in simulation mode?

Nirusu (changes requested):

samples/graphene-hello/preload/README.md
m1ghtym0: > Graphene specifically requires SGX-enabled hardware. You cannot run Graphene applications in simulation mode. And if you already run the application on an SGX-enabled system, why specifically run the Coordinator in simulation mode?

You can run Graphene without SGX, but we're currently not doing this in our Makefile.
Most people probably won't have SGX enabled on their machines (and if they don't run in Azure, they won't have a DCAP infrastructure to verify quotes), hence, the incentive for OE_SIMULATION makes sense to me. We can argue that people TRYING our examples are more likely to have access to some SGX machine.

> however for the samples I think it is important to demonstrate that with the UniqueID (or other parameters) pinned in the manifest, you cannot simply modify your application as you please. This only works properly in non-simulation mode.

I see your point. For me, it's a compromise between keeping the bar to interact with our samples as low as possible and actually having meaningful samples.
I agree with @Nirusu that the current setup of Coordinator in simulation mode and Graphene in SGX mode does not make sense because we get neither of those traits.
I'd say we remove the `OE_SIMULATION` because building the Graphene enclave requires a working Graphene environment which most people will have as their "Cloud Development Setup". It might be useful for us to have something similar in our docs: https://graphene.readthedocs.io/en/latest/cloud-deployment.html

katexochen: Maybe we should provide an overview of the examples sorted by their requirements with categories like "no-sgx", "sgx-vm", "sgx-cluster"?
I like the idea of having some "prerequisite setup" tutorials as @m1ghtym0 suggested.

Nirusu: > You can run Graphene without SGX, but we're currently not doing this in our Makefile.

Ah, now this was a bad oversimplification on my part. Yes, you can actually run Graphene without SGX, but you need to have build the entire tool set once without SGX=1 initially (otherwise you're missing the non-SGX pal-loader by default) and we would need to adjust our Makefile.

What I was meant to say here is that you cannot simply interchange it _that_ easily as with our tools by just running something prebuilt in simulation mode. It has to be already considered for when setting up / building Graphene, thus you would go down an entirely different path (non-SGX build) during the initial approach anyway.

> For me, it's a compromise between keeping the bar to interact with our samples as low as possible and actually having meaningful samples.

I agree here. But I think it is important that we do not simply append OE_SIMULATION=1 as default, as we basically bypass security here. We basically do not even run correctly on SGX, we just simulate it. Now I get that the DCAP infrastructure is still immature, users might likely not have the required environment here even if their hardware supports SGX etc., but I think we should aim for two things:

- Whenever we can fix certain situations in our tooling without a security risk, we should do it in tooling. 
- Above that, we should educate the user here about any insecurities for going down "the easy way"

We advertise Marblerun as a SGX-cluster solution. I do not think it would make sense to assume by default that we do not care about SGX anymore just to make them errors go away. I agree that for initial setup in a testing environment, it properly does not matter to do everything 100% correctly. But then the user should know about it, or have to follow a (slightly) different tutorial path. But we should not go for "insecure by default, afterwards switch to secure whenever you're done".

> Maybe we should provide an overview of the examples sorted by their requirements with categories like "no-sgx", "sgx-vm", "sgx-cluster"?

I think @katexochen 's suggestion to have different categories here might be good, though I hope we can keep this short but sweet. Otherwise, I feat that it could result in TL;DR if it gets to long. 

But I think we need to educate the user somehow on the current landscape of SGX, DCAP, what is/should be correct and what not, but allows them testing at least 90% of the feature set and UX. After reading tutorials & running the samples, they should be at least somewhat confident in using our tools.

I think if you go for Marblerun, you properly care about confidential computing in someway. Now, if you bypass everything, you should at least know that this is not quite intended to work this way it is supposed too, but that we just choose it for initial simplicity. 

Therefore, please do not set this as default, but rather as an alternative way :)

katexochen: > Therefore, please do not set this as default, but rather as an alternative way :)

just to clarify, this is nothing I introduced, the sample previous used the simulation mode, since it referred to the [BUILD.md](https://github.com/edgelesssys/marblerun/blob/master/BUILD.md#run-the-coordinator). I just copied the command from there.

In my opinion, tutorials should be as linear as possible, therefore I wouldn't add alternative ways. They should describe a single exemplary scenario.

I've deleted the OE_SIMULATION=1 in the Graphene examples and added a line mentioning that SGX is needed. I've also added an explanation and security note to the examples that don't use SGX (the C++ and Hello World example).

Nirusu: > just to clarify, this is nothing I introduced, the sample previous used the simulation mode, since it referred to the BUILD.md. I just copied the command from there.

Yup, I get that. But I think the BUILD.md [might need a note as we do for our quickstart guide](https://www.marblerun.sh/docs/getting-started/quickstart/), instead of just starting without any mention.

Though then it is a BUILD guide, and many of our tutorials assume starting with the prebuilt binaries. 

I think there's still plenty of room for improvement in our documentation. Or for the SGX-world in general, it is still confusing in most tools. And I think we can try to do better.

> In my opinion, tutorials should be as linear as possible, therefore I wouldn't add alternative ways.

I do not think it's too bad if done well. Especially for the libOS samples (Graphene, Occlum), their guides do something similar. Graphene has seperate non-SGX, SGX steps. [Occlum mentions shortly you can run this in simulation mode](https://github.com/occlum/occlum#hello-occlum) for non-SGX platforms, which is similar on how we do it on our website.

The note you added seems okay for me for this PR, I would be okay with it! 

For the future, I think it would be great if we could:

- Develop a small program which determines the current system's SGX capabilities (Does CPU support it? Is it set up correctly? Does DCAP work?)
- Give users a simple answer, give them a path for a "Getting Started" guide and our samples
- Expect requirements / limitations of possible paths
- Keep things as streamlined as possible, but make seperate trails for cases where we really cannot streamline it in a way it makes sense

And I think adding OE_SIMULATION=1 in a non-ideal scenario does not require too much alternative ways.

But yeah, for the sake of the current PR, I think such a small note for the "simple" non-libOS samples is okay, and for the libOS samples we could assume they require SGX hardware or we can assume that the user should know that many tools allow switching you between non-SGX/simulation mode and SGX-mode, because at this point they properly already went through their documentation to build their application.

m1ghtym0:
Changes lgtm:-)
I'll take a look at why the build is failing since the last edgelessrt release.

README.md
```diff
@@ -59,6 +59,7 @@ We provide basic examples on how to build confidential apps with Marblerun.
 * See [helloc++](samples/helloc++) for an example in C++.
 * See [graphene-hello](samples/graphene-hello) for an example using Graphene.
 * See [graphene-nginx](samples/graphene-nginx) for an example of converting an existing Graphene application to a Marble.
+* See [graphene-redis](samples/graphene-redis) for a distributed redis example using Graphene.
```
Nirusu, line 62:
```suggestion
* See [graphene-redis](samples/graphene-redis) for a distributed Redis example using Graphene.
```

And subsequently, everywhere else.

samples/graphene-hello/preload/README.md
```diff
@@ -1,49 +1,42 @@
...
-- EDG_COORDINATOR_CLIENT_ADDR=localhost:4433
-- EDG_COORDINATOR_DNS_NAMES=localhost
-- EDG_COORDINATOR_SEAL_DIR=$PWD
+Next, use the `erthost` command to start the Coordinator in a local enclave:
+```sh
+erthost ../../../build/coordinator-enclave.signed
+```
 
-Once the [Coordinator instance is running](../../BUILD.md#run-the-coordinator), upload the manifest to the Coordinator:
+The Coordinator exposes two APIs, a client API to instruct the Coordinator (port 4433) and a mesh API to communicate with your Marble (port 2001).
```
Nirusu, line 27: I think the wording "The coordinator exposes... to instruct the coordinator" is a bit confusing. 

Maybe we can reword this a bit, highlight it's a REST API which an admin can use to manage the coordinator (e.g. setting a manifest) or something like that. Opinions?

samples/graphene-hello/preload/README.md
```diff
@@ -1,49 +1,42 @@
...
 
+Once the Coordinator instance is running, you can upload the mainfest to the Coordinators client API:
```
Nirusu, line 29:
```suggestion
Once the Coordinator instance is running, you can upload the mainfest to the Coordinator's client API:
```

samples/graphene-hello/preload/README.md
```diff
@@ -1,49 +1,42 @@
...
 ```
 curl -k --data-binary @manifest.json https://localhost:4433/manifest
 ```
 
-Now we can run our application:
+To run the application, you need to set some environment variables. The Marbles type is defined in the `manifest.json`. In this example, the manifest defines a single Marble, which is called "hello". The Marbles DNS name and the Coordinators address are used to establish a connection between the Coordinators mesh API and the Marble, and the UUID file stores a unique ID that enables a restart of the application.
```
Nirusu, line 34:
```suggestion
To run the application, you need to set some environment variables. The Marble's type is defined in the `manifest.json`. In this example, the manifest defines a single Marble, which is called "hello". The Marble's DNS name and the Coordinator's address are used to establish a connection between the Coordinator's mesh API and the Marble. At last, the UUID file stores a unique ID that enables a restart of the application.
```

or:

```suggestion
To run the application, you need to set some environment variables. The type of the Marble is defined in the `manifest.json`. In this example, the manifest defines a single Marble, which is called "hello". The Marble's DNS name and the Coordinator's address are used to establish a connection between the Coordinator's mesh API and the Marble. At last, the UUID file stores a unique ID that enables a restart of the application.
```

Properly would opt for the latter to not sound too repetitive?

samples/graphene-hello/preload/README.md
```diff
@@ -1,49 +1,42 @@
...
-EDG_MARBLE_COORDINATOR_ADDR=localhost:2001 EDG_MARBLE_TYPE=hello EDG_MARBLE_UUID_FILE=uuid EDG_MARBLE_DNS_NAMES=localhost make run
-```
-
-## Troubleshooting
-If you receive the following error message on launch:
-
+EDG_MARBLE_TYPE=hello \
+EDG_MARBLE_COORDINATOR_ADDR=localhost:2001 \
+EDG_MARBLE_UUID_FILE=uuid \
+EDG_MARBLE_DNS_NAMES=localhost \
```
Nirusu, line 40: Haven't tested it, but can we define `loader.insecure__use_host_env = 1`, yet still define additional environment variables with `loader.env.[ENVIRON] = "[VALUE]"`?

If yes, I think we should hide those in the Graphene manifest.

samples/graphene-hello/spawn/README.md
```diff
@@ -1,48 +1,41 @@
...
-- EDG_COORDINATOR_SEAL_DIR=$PWD
+Next, use the `erthost` command to start the Coordinator in a local enclave:
+```sh
+erthost ../../../build/coordinator-enclave.signed
+```
 
-Once the [Coordinator instance is running](../../BUILD.md#run-the-coordinator), upload the manifest to the Coordinator:
+The Coordinator exposes two APIs, a client API to instruct the Coordinator (port 4433) and a mesh API to communicate with your Marble (port 2001).
 
+Once the Coordinator instance is running, you can upload the mainfest to the Coordinators client API:
```
Nirusu, line 28:
```suggestion
Once the Coordinator instance is running, you can upload the mainfest to the Coordinator's client API:
```

Same as above. Maybe if you opt for the change, please manually add them here too. Don't want to add too many comments here with the same thing.

samples/graphene-nginx/README.md
```diff
@@ -1,31 +1,38 @@
...
 ```
 
-We assume that the Coordinator is run with the following environment variables:
-- EDG_COORDINATOR_MESH_ADDR=localhost:2001
-- EDG_COORDINATOR_CLIENT_ADDR=localhost:4433
-- EDG_COORDINATOR_DNS_NAMES=localhost
-- EDG_COORDINATOR_SEAL_DIR=$PWD
-
-Once the [Coordinator instance is running](../../BUILD.md#run-the-coordinator), upload the manifest to the Coordinator:
+Start the Coordinator in a local simulated enclave:
```
Nirusu, line 17:
```suggestion
Start the Coordinator in a SGX enclave:
```

Not simulated anymore.

samples/graphene-redis/README.md
```diff
@@ -1,21 +1,21 @@
-# Graphene redis sample
+# Graphene redis example
```
Nirusu, line 1: Also here, please capitalize Redis everywhere.

samples/helloc++/README.md
```diff
@@ -1,12 +1,14 @@
-# C++ sample
-This sample shows how to build a confidential C++ application and run it in Marblerun. This should serve as a blueprint for making existing applications Marblerun-ready or creating new ones.
+# How to create a C++ Marble
+This example shows how to build a confidential C++ application and run it in Marblerun. This can serve you as a blueprint for making existing applications Marblerun-ready or creating new [Marbles](https://www.marblerun.sh/docs/getting-started/marbles/). If you haven't already, [setup Marblerun](../../BUILD.md#build) to get ready.
+
+You can run this example on any hardware, since the enclave is only simulated (with the `OE_SIMULATION=1` environment variable). This might help you to get started with with the development of confidential apps. However, please notice that this mode does not provide any safety. Detailed information on how to develope secure Marbles can be found in [Marbleruns documentation](https://www.marblerun.sh/docs/tasks/add-service/).
```
Nirusu, line 4: Okay, now I think this actually is a bit confusing if we keep this at default for some samples, but for others not. 

I still think that non-simulation should be default, but we should use the note to highlight that you can enable it. Not vice versa? Also, it's now kind of inconsistent between the samples.

But you decide.

samples/helloc++/README.md
```diff
@@ -15,26 +17,32 @@ cmake ..
...
-- EDG_COORDINATOR_SEAL_DIR=$PWD
+Next, use the `erthost` command to start the Coordinator in a local simulated enclave:
+```sh
+OE_SIMULATION=1 erthost ../../../build/coordinator-enclave.signed
+```
+
+The Coordinator exposes two APIs, a client API to instruct the Coordinator (port 4433) and a mesh API to communicate with your Marble (port 2001).
 
-Once the [Coordinator instance is running](../../BUILD.md#run-the-coordinator), upload the manifest to the Coordinator:
+Once the Coordinator instance is running, you can upload the mainfest to the Coordinators client API:
```
Nirusu, line 35:
```suggestion
Once the Coordinator instance is running, you can upload the mainfest to the Coordinator's client API:
```

samples/helloc++/README.md
```diff
@@ -15,26 +17,32 @@ cmake ..
...
 ```sh
 curl -k --data-binary @manifest.json https://localhost:4433/manifest
 ```
 
-Now we can run our application:
+To run the application, you need to set some environment variables. The Marbles type is defined in the `manifest.json`. In this example, the manifest defines a single Marble, which is called "hello". The Marbles DNS name and the Coordinators address are used to establish a connection between the Coordinators mesh API and the Marble, and the UUID file stores a unique ID that enables a restart of the application.
```
Nirusu, line 40: Same change as above?

samples/graphene-hello/README.md
```diff
@@ -1,15 +1,15 @@
...
-This sample shows how to run a [Graphene](https://github.com/oscarlab/graphene) application in Marblerun.
+# Graphene "Hello World!" example
+This example shows how to run a [Graphene](https://github.com/oscarlab/graphene) application in Marblerun.
 
-In essence, you have to add the `premain` process either by manually adding the Marblerun premain process as the entry point for your application (see: `spawn`), or by automatically injecting it as a preloaded shared library via Graphene's LD_PRELOAD feature (see: `preload`).
+In essence, you have to add the `premain` process either by manually adding the Marblerun premain process as the entry point for your application (see: [`spawn`](spawn)), or by automatically injecting it as a preloaded shared library via Graphene's LD_PRELOAD feature (see: [`preload`](preload)).
 
-`premain` will contact the Coordinator, set up the environment, and run the actual application. See the commented hello.manifest.template in the corresponding folder for details.
+`premain` will contact the Coordinator, set up the environment, and run the actual application. See the commented `hello.manifest.template` in the corresponding folder for details.
```
Nirusu, line 6:
```suggestion
`premain` will contact the Coordinator, set up the environment, and run the actual application. See the commented [`hello.manifest.template`](hello.manifest.template) in the corresponding folder for details.
```

Hope this links works locally.

samples/helloc++/README.md
```diff
@@ -15,26 +17,32 @@ cmake ..
...
 ```sh
-EDG_MARBLE_COORDINATOR_ADDR=localhost:2001 EDG_MARBLE_TYPE=hello EDG_MARBLE_UUID_FILE=$PWD/uuid EDG_MARBLE_DNS_NAMES=localhost erthost enclave.signed
+EDG_MARBLE_TYPE=hello \
+EDG_MARBLE_COORDINATOR_ADDR=localhost:2001 \
+EDG_MARBLE_UUID_FILE=$PWD/uuid \
+EDG_MARBLE_DNS_NAMES=localhost \
+erthost build/enclave.signed
 ```
+The app prints a "Hello world!" followd by the commandline arguments that are defined in the `manifest.json`.
```
Nirusu, line 48:
```suggestion
The app prints a "Hello world!" followed by the command line arguments defined in `manifest.json`.
```

samples/helloworld/README.md
```diff
@@ -1,36 +1,39 @@
...
 curl -k --data-binary @manifest.json https://localhost:4433/manifest
 ```
 
-Now we can run our application:
+Finaly, you can run the helloworld Marble (or whatever Marble you just created) with the `ego marblerun` command. You just need to set `EDG_MARBLE_TYPE` to a Marble that was defined in the `manifest.json`. In this example, the manifest defines a single Marble, which is called "hello".
 ```sh
-EDG_MARBLE_COORDINATOR_ADDR=localhost:2001 EDG_MARBLE_TYPE=hello EDG_MARBLE_UUID_FILE=$PWD/uuid EDG_MARBLE_DNS_NAMES=localhost ego marblerun helloworld
+EDG_MARBLE_TYPE=hello ego marblerun helloworld
 ```
+Marblerun starts the Marble and connects it to the mesh API of the Coordinator.
```
Nirusu, line 34:
```suggestion
EGo starts the Marble and connects it to the mesh API of the Coordinator.
```

I think? EGo calls the Premain from Marblerun.

Nirusu (commented):
Some suggestions I have. Feel free to apply or discuss what you think is worth changing, or what you disagree with.

samples/graphene-hello/README.md
katexochen: This won't work, since the `graphene-hello` folder does not contain such a template. In the subfolders `preload` and `spawn`, where such a file exists, the link to the file already exists :wink:

samples/graphene-hello/preload/README.md
katexochen: I'm not familiar with Graphene, so it would be nice if you could test this and submit a suggestion.

samples/graphene-hello/README.md
Nirusu: Oh, didn't see this was actually the README.md from the root sample directory. My bad, that's fine then!

samples/graphene-hello/preload/README.md
m1ghtym0: I'd keep them in the user's environment. The `"Don't use this in production!"` message is not ideal, but moving them into the Graphene manifest essentially makes them constants defeating the env-var purpose entirely.

Nirusu (approved):

m1ghtym0 (approved):

PR #153, opened 2021-05-03
Replace spawn with execve in Graphene premain & remove `preload` method

### LINKED ISSUE #151: [premain-Graphene] `posix_spawn` is slow and could be replaced with `execve`
I just noticed that `premain-graphene` code uses `posix_spawn()`. On Linux-x64, this is implemented via the combination of `vfork() + execve()`.

In Graphene, `vfork()` is a slow system call because it must create a new SGX enclave and securely copy the complete enclave-memory state from the parent enclave to the new child enclave.

However, `execve()` in Graphene is a very fast system call because there is no need to create a new enclave -- the parent enclave "becomes" the child enclave.

So in the below line you could actually use `execve` instead of `posix_spawn`: https://github.com/edgelesssys/marblerun/blob/211bd99818dee8deeb69da435fbe58f6d9854c56/cmd/premain-graphene/main.go#L44

And then everything would be fast, and you won't need two separate ways of running the pre-main functionality of Marbles.

This will also simplify your examples and the tutorials and the README: https://github.com/edgelesssys/marblerun/tree/master/samples/graphene-hello

You will get rid of what you call the `preload` method and only keep the `spawn` method which will not have the disadvantage of being slow.

Nirusu:
This incorporates the changes suggested by @dimakuv suggested in #151. 

(Closes #151.)


cmd/premain-graphene/main.go
```diff
@@ -40,11 +39,10 @@ func main() {
 	argv := toCArray(os.Args)
 	envp := toCArray(os.Environ())
 
-	// spawn service
-	if res := C.posix_spawn(nil, C.CString(service), nil, nil, &argv[0], &envp[0]); res != 0 {
+	// launch service
+	if res := C.execve(C.CString(service), &argv[0], &envp[0]); res != 0 {
```
thomasten, line 43: Can you try if https://golang.org/pkg/syscall/#Exec works here?

cli/cmd/graphenePrepare.go
```diff
@@ -36,14 +33,7 @@ const longDescription = `Modifies a Graphene manifest for use with Marblerun.
...
 
-The first parameter of this command is either 'spawn' or 'preload'.
-
-'spawn': Replace the entrypoint of your application with Marblerun's premain. Dedicates argv provisioning to Marblerun's manifest, but takes longer to load.
-
-'preload': Loads Marblerun's premain as a shared library via LD_PRELOAD and keeps your original entrypoint intact.
-This feature delegates argv provisioning to Graphene, making Marblerun unable to supply its own arguments via the Marblerun manifest, but keeps better compability with existing Graphene applications and leads to faster load times.
-
-For more information about both modes, consult the documentation: https://www.marblerun.sh/docs/tasks/build-service-graphene/
+For more information about the requirements and  changes performed, consult the documentation: https://www.marblerun.sh/docs/tasks/build-service-graphene/
```
thomasten, line 36:
```suggestion
For more information about the requirements and  changes performed, consult the documentation: https://www.marblerun.sh/docs/building-services/graphene/
```

cli/cmd/graphenePrepare.go
```diff
@@ -36,14 +33,7 @@ const longDescription = `Modifies a Graphene manifest for use with Marblerun.
...
 
 The second parameter of this command is the path of the Graphene manifest template you want to modify.
```
thomasten, line 38:
```suggestion
The parameter of this command is the path of the Graphene manifest template you want to modify.
```

cli/cmd/graphenePrepare.go
```diff
@@ -127,47 +102,28 @@ func parseTreeForChanges(tree *toml.Tree, mode mode) (map[string]interface{}, ma
...
-		}
-	}
-
-	if original["libos.entrypoint"].(string) == "file:"+premainNameSpawn || original["sgx.trusted_files.marblerun_premain"] != nil || original["sgx.allowed_files.marblerun_uuid"] != nil {
+	if original["libos.entrypoint"].(string) == "file:"+premainName || original["sgx.trusted_files.marblerun_premain"] != nil || original["sgx.allowed_files.marblerun_uuid"] != nil {
+		color.Yellow("The supplied manifest already contains changes for Marblerun. Have you selected the correct file?")
 		return nil, nil, errors.New("manifest already contains Marblerun changes")
 	}
 
 	// Add changes to entry point depending on mode
```
thomasten, line 110:
```suggestion
```

CMakeLists.txt
```diff
@@ -127,4 +127,3 @@ add_custom_target(sign-marble-test ALL DEPENDS marble-test-enclave.signed marble
 #
 
 add_custom_target(premain-graphene ALL ertgo build ${TRIMPATH} -buildmode=pie ${CMAKE_SOURCE_DIR}/cmd/premain-graphene)
-add_custom_target(premain-graphene-so ALL ertgo build ${TRIMPATH} -buildmode=c-shared -o premain-graphene.so ${CMAKE_SOURCE_DIR}/cmd/premain-graphene-so)
```
thomasten, line 130: also remove artifact from unittests.yml

thomasten (changes requested):

cmd/premain-graphene/main.go
Nirusu: Works, so no more CGo needed here. Great!

dimakuv (approved):
Took a brief look at the PR. Looks legitimate to me. Thanks for this!

thomasten (approved):

PR #154, opened 2021-05-04
Bump version to 0.3.2

m1ghtym0:


thomasten:
How should we handle
https://github.com/edgelesssys/marblerun/blob/master/cmd/coordinator/run.go#L24
https://github.com/edgelesssys/marblerun/blob/master/cli/cmd/version.go#L12
?
Keep updating it?
Replacing it with v0.0.0-dev or the like?

m1ghtym0:
> How should we handle
> https://github.com/edgelesssys/marblerun/blob/master/cmd/coordinator/run.go#L24
> https://github.com/edgelesssys/marblerun/blob/master/cli/cmd/version.go#L12
> ?
> Keep updating it?
> Replacing it with v0.0.0-dev or the like?

Good catch! Setting it to v0.0.0-dev makes sense to me. @daniel-weisse what do you say?

daniel-weisse:
I dont think replacing it with v0.0.0-dev is the best idea. Might confuse someone into thinking the version is way lower than what it should be.
Maybe just setting it to something like `dev-build` would work

cli/cmd/graphenePrepare_test.go
```diff
@@ -142,5 +142,5 @@ func TestDownloadPremain(t *testing.T) {
 
 	// We should have one download here
 	info := httpmock.GetCallCountInfo()
-	assert.Equal(1, info[`GET =~^https://github\.com/edgelesssys/marblerun/releases/download/v[0-9\.]*/premain-graphene`])
+	assert.Equal(1, info[`GET =~^https://github\.com/edgelesssys/marblerun/releases/download/.*/premain-graphene`])
```
Nirusu, line 145: Mhh, I don't know about `dev-build`.

Personally, having the version encoded in a sane way allows us to unit test download URLs or anything where we dynamically parse the version into.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.11)
 
-project(marblerun VERSION 0.3.1)
+project(marblerun VERSION 0.3.2)
 execute_process(COMMAND bash "-c" "git rev-parse HEAD | tr -d '\n'" OUTPUT_VARIABLE GIT_COMMIT)
```
Nirusu, line 4: The GIT_COMMIT injection is also kind of broken right now, I recently noticed.

It only gets/updates the GIT_COMMIT when it creates the build files (`cmake ..`), but not actually during build times. Meaning you end up with an outdated git commit string when you don't rebuild your build directory every single time.

m1ghtym0: That's correct. However, this was best effort. If there is a better way to do it with cmake we should fix it?

cli/cmd/graphenePrepare_test.go
m1ghtym0: > Mhh, I don't know about `dev-build`.

I don't have any strong opinion here. It is a placeholder value for a version string that is replaced at build time.
So I'm open to suggestions. I see @daniel-weisse's point that `v0.0.0-dev` could confuse people. But if there is no better option though, I'm fine with `v0.0.0` or `v0.0.0-dev` as well. It's not that big of a deal IMO.

> Personally, having the version encoded in a sane way allows us to unit test download URLs or anything where we dynamically parse the version into.

Sure. Ideally, that's the case. 

> This change here basically breaks the test. It's intend was to check if it sets a plausible URL, now it straight out just accepts every sequence of characters, which is not great and makes the test kinda obsolete.

It accepts every type of version string, the URL is still intact. We're basically testing the parsing of placeholder value here, so I could live with that.

If there is no objection from @daniel-weisse I'll go with the `v0.0.0` string and add a comment.

daniel-weisse: Sounds good to me

Nirusu (approved):

PR #162, opened 2021-05-17
AB#901 Update commit hash on every build

daniel-weisse:
Currently, updating the commit hash, as displayed by `marblerun version` and by the Coordinator on start, requires setting up a new build environment.
This PR changes this by adding a new step to `CMakeLists.txt` to generate a file (`util/version.go`) containing version and commit information.

Things to think about before merging:
I am unaware of a method to keep a file in the github repo while also ignoring any local changes to it.

`version.go` will get changed every time someone runs build after creating a new commit. It will then show up as part of the files not staged for commit.
To avoid this one can remove the file from the worktree: `git update-index --skip-worktree util/version.go`.
However this is local only, meaning everyone would have to set it up themselves on their machine.

Another option is removing `util/version.go` from the repo and relying on `.gitignore` to not add it back once it gets created during build.
This would not require any additional steps to avoid pushing an "updated" version file.
The downside would be that a newly cloned repo will report errors in the files relying on `util/version.go`, since it has not yet been created. Only after the first time `make` is ran this will be fixed.


thomasten:
Hm, I'd prefer to stick to -ldflags -X. We might invoke a short script instead of `go build` that gets the current hash and then calls `go build`. Maybe you can also achieve this with command substitution instead of a script (https://stackoverflow.com/a/48639184).

daniel-weisse:
Could be worth a try.
I dont think there is a way to generate a cmake variable at build time using custom commands, so we would need to include a small script for each target requiring the commit hash. We could create a variable local to that script and then maybe use it with `-ldflags`

thomasten:
> script for each target

maybe this can be the same script for each target, with parameters

thomasten (approved):
Interesting. I thought about just using a shell script, but this seems to be all right, too.

PR #164, opened 2021-05-17
add ttls server side

3u13r:


test/manifests.go
```diff
@@ -156,6 +161,13 @@ const ManifestJSON string = `{
 					"Port": "40000",
 					"Addr": "example.com"
 				}
+			],
+			"Incoming": [
+				{
+					"Port": "8080",
+					"Cert": "{{ pem .Secrets.cert_shared.Cert }}",
+					"Key": "{{ pem .Secrets.cert_shared.Private }}"
```
thomasten, line 169:
```suggestion
					"Cert": "cert_shared"
```
We should be able to get the secret by its name.

coordinator/core/marbleapi.go
```diff
@@ -315,33 +315,68 @@ func (c *Core) generateMarbleAuthSecrets(req *rpc.ActivationReq, marbleUUID uuid
...
-	pemClientCert := pem.Block{Type: "CERTIFICATE", Bytes: secrets.MarbleCert.Cert.Raw}
+	pemClientCert := pem.Block{Type: "CERTIFICATE", Bytes: specialSecrets.MarbleCert.Cert.Raw}
 	stringClientCert := string(pem.EncodeToMemory(&pemClientCert))
 
-	pemClientKey := pem.Block{Type: "PRIVATE KEY", Bytes: secrets.MarbleCert.Private}
+	pemClientKey := pem.Block{Type: "PRIVATE KEY", Bytes: specialSecrets.MarbleCert.Private}
 	stringClientKey := string(pem.EncodeToMemory(&pemClientKey))
 
 	for _, tag := range marble.TLS {
+		connConf := make(map[string]string)
```
thomasten, line 344: I'd move this back in the inner loop(s) because otherwise you have to be careful that one config doesn't contain a value from another config. E.g., incoming user-defined may have a cacrt.

coordinator/manifest/manifest.go
```diff
@@ -114,6 +116,29 @@ func (m Manifest) Check(ctx context.Context, zaplogger *zap.Logger) error {
...
+		for _, tag := range marble.TLS {
+			if _, ok := m.TLS[tag]; !ok {
+				return fmt.Errorf("manifest misses TLS entry for %s", tag)
+			}
+		}
+	}
+	for key, TLStag := range m.TLS {
+		for _, entry := range TLStag.Incoming {
+			if entry.Port == "" {
+				return fmt.Errorf("manifest misses Port in Tls.Incoming.%s", key)
```
thomasten, line 128:
```suggestion
				return fmt.Errorf("manifest misses Port in TLS.Incoming.%s", key)
```

coordinator/core/marbleapi.go
```diff
@@ -315,33 +315,68 @@ func (c *Core) generateMarbleAuthSecrets(req *rpc.ActivationReq, marbleUUID uuid
...
+					return err
+				}
+				connConf["clicert"] = cliCrt
+				connConf["clientAuth"] = "disabled"
+
+			} else {
+				connConf["cacrt"] = stringCaCert
+				connConf["clicert"] = stringClientCert
+				connConf["clikey"] = stringClientKey
+				connConf["clientAuth"] = "enabled"
```
thomasten, line 373: should be bool

thomasten (changes requested):

thomasten (approved):

PR #170, opened 2021-05-31
AB#967 Add a store interface

daniel-weisse:
This PR adds the store interface to the coordinator.
This allows different implementations and approaches of storing and persisting data used by the coordinator.

Currently only the standard implementation in `coordinator/core/stdcore.go` is used. This implementation behaves the same as the old implementation of the core.

The `Store` interface implements 5 methods:
* `Get`: sends a request in the form of a string to the store and receives a byte array and/or error as response
* `Put`: send a request to save a byte array in the store referenced by a string. This data can then be requested using `Get`.
* `SealState`: encrypts and persists the data of the store to disk
* `LoadState`: decrypts data saved by `SealState` and loads them into the store
* `SetEncryptionKey`: sets the encryption key for `LoadState` and `SealState`

Using the methods implemented in `storewrapper` the different methods of `Store` are abstracted into more usable ways by taking care of unit conversions etc. For example when requesting the activations of a specific marble we can just call `getActivations(marbleType)` and get the activations as a `uint`.

### DIFF
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -18,7 +18,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
-	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
@@ -38,34 +37,15 @@ import (
 
 // Core implements the core logic of the Coordinator
 type Core struct {
-	rootCert          *x509.Certificate
-	intermediateCert  *x509.Certificate
-	adminCerts        []*x509.Certificate
-	quote             []byte
-	rootPrivK         *ecdsa.PrivateKey
-	intermediatePrivK *ecdsa.PrivateKey
-	sealer            Sealer
-	recovery          recovery.Recovery
-	manifest          manifest.Manifest
-	rawManifest       []byte
-	updateManifest    manifest.Manifest
-	rawUpdateManifest []byte
-	secrets           map[string]manifest.Secret
-	state             state
-	store             Store
-	qv                quote.Validator
-	qi                quote.Issuer
-	activations       map[string]uint
-	mux               sync.Mutex
-	zaplogger         *zap.Logger
-}
-
-// Store is the interface for state transactions and persistance
-type Store interface {
-	Get(string) ([]byte, error)
-	Put(string, []byte) error
-	SealState() error
-	LoadState() error
+	quote          []byte
+	recovery       recovery.Recovery
+	manifest       manifest.Manifest
+	updateManifest manifest.Manifest
+	store          *storeWrapper
+	qv             quote.Validator
+	qi             quote.Issuer
+	mux            sync.Mutex
+	zaplogger      *zap.Logger
 }
 
 // The sequence of states a Coordinator may be in
@@ -79,17 +59,12 @@ const (
 	stateMax
 )
 
-// sealedState represents the state information, required for persistence, that gets sealed to the filesystem
-type sealedState struct {
-	RootPrivK           []byte
-	IntermediatePrivK   []byte
-	RawManifest         []byte
-	RawUpdateManifest   []byte
-	RawRootCert         []byte
-	RawIntermediateCert []byte
-	Secrets             map[string]manifest.Secret
-	State               state
-	Activations         map[string]uint
+// marblerunUser represents a privileged user of Marblerun
+type marblerunUser struct {
+	// name is the username
+	name string
+	// certificate is the users certificate, used for authentication
+	certificate *x509.Certificate
 }
 
 // coordinatorName is the name of the Coordinator. It is used as CN of the root certificate.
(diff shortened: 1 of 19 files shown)


coordinator/core/storewrapper.go
```diff
@@ -0,0 +1,262 @@
...
+	requestCert        string = "certificate"
+	requestManifest    string = "manifest"
+	requestPrivKey     string = "privateKey"
+	requestSecret      string = "secret"
+	requestState       string = "state"
+	requestUser        string = "user"
+)
+
+// Store is the interface for state transactions and persistance
+type Store interface {
```
m1ghtym0, line 30: Should `Store` be in a separate module?
Not sure what's best practice here @thomasten?
Another thought: Should `Get` and `Put` have the prefix/family as an argument instead of joining them in the wrapper?
This would leave it to the plugins to handle the prefix/family. Some storage backend might support this for performance reasons.

daniel-weisse: I think it is better to have `Get` and `Put` only take a single argument as the key. This leaves the interface as a simple key-value store.
I dont think the performance impact of joining or splitting a short string will be noticeable, even with thousands of requests per second.

coordinator/core/core.go
```diff
@@ -92,67 +76,113 @@ const coordinatorIntermediateName string = "Marblerun Coordinator - Intermediate
...
+		}
+		if err := c.store.putPrivK("root", rootPrivK); err != nil {
+			return nil, err
+		}
+		if err := c.store.putPrivK("intermediate", intermediatePrivK); err != nil {
+			return nil, err
+		}
 		c.advanceState(stateRecovery)
-	} else if rootCert == nil {
+	} else if _, err := c.store.getCertificate("root"); isStoreValueUnsetError(err) {
```
thomasten, line 149: what if this returns another error?

coordinator/core/core.go
```diff
@@ -92,67 +76,113 @@ const coordinatorIntermediateName string = "Marblerun Coordinator - Intermediate
...
+		if err := c.store.putCertificate("intermediate", intermediateCert); err != nil {
+			return nil, err
+		}
+		if err := c.store.putPrivK("root", rootPrivK); err != nil {
+			return nil, err
+		}
+		if err := c.store.putPrivK("intermediate", intermediatePrivK); err != nil {
+			return nil, err
+		}
 		c.advanceState(stateAcceptingManifest)
```
thomasten, line 172: reduce duplication

thomasten (changes requested):
I think this is the right direction. Next step should be adding transactions before adding new features.
@daniel-weisse @m1ghtym0 pls check if https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/997 makes sense to do next.
I didn't do a complete review. Either @m1ghtym0 (if not already done) or @Nirusu should do this before we merge this.

coordinator/core/storewrapper.go
thomasten: I agree with @m1ghtym0 regarding separate module.
@daniel-weisse It's not about string operation cost, but KV stores that support grouping keys. However, I think this is negligible for our use case. I don't have a strong opinion, only a slight preference to keep it simple.

coordinator/core/storewrapper.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	activations, err := strconv.ParseUint(string(rawActivations), 16, 64)
+	return uint(activations), err
+}
+
+// putActivations saves activations of a given Marble to store
+func (s *storeWrapper) putActivations(marbleType string, activations uint) error {
+	request := strings.Join([]string{requestActivations, marbleType}, ":")
+	rawActivations := []byte(strconv.FormatUint(uint64(activations), 16))
```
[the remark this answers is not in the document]
daniel-weisse: No specific reason to convert to hex, other than maybe saving a very small amount of space.
We convert to string at this point to easily save the data as `[]byte`. Not choosing hex would result in strings with slightly more characters for large number of activations.

coordinator/core/storewrapper.go
```diff
@@ -0,0 +1,260 @@
...
+	secretMap := map[string]manifest.Secret{}
+
+	rawManifest, err := s.getRawManifest("main")
+	if err != nil {
+		return nil, err
+	}
+	var manifest manifest.Manifest
+	if err := json.Unmarshal(rawManifest, &manifest); err != nil {
+		return nil, err
+	}
```
m1ghtym0, line 171:
```suggestion
	manifest, err := s.getManifest("main")
	if err != nil {
		return nil, err
	}
```

coordinator/core/core.go
```diff
@@ -38,25 +39,15 @@ import (
...
-	state             state
-	qv                quote.Validator
-	qi                quote.Issuer
-	activations       map[string]uint
-	mux               sync.Mutex
-	zaplogger         *zap.Logger
+	quote          []byte
+	recovery       recovery.Recovery
+	manifest       manifest.Manifest
+	updateManifest manifest.Manifest
```
m1ghtym0, line 45: Why are we keeping `manifest` and `updateManifest` in the Core? They just keep duplicate values to the ones in the store, no?

m1ghtym0 (changes requested):

coordinator/core/core.go
daniel-weisse: Its more of a convenience thing than anything else. Since the manifest does not change after it has been set once, it is kept as a `manifest.Manifest` object in memory. This removes the need to load the manifest from the store every time we want to iterate over values set by the manifest (for example when generating secrets for a marble).

We can of course change this to instead fetch it just in time using the `storeWrapper` `getManfiest` method

m1ghtym0: Yeah, I'd keep it simple and omit caching for now. Considering future changes and potentially a distributed Coordinator we should refrain from special cases at this point.

thomasten: It may be best to parse the manifest once and insert the values into the store. But does not need to happen within this PR.

thomasten: We can then also get rid of the update manifest.

m1ghtym0 (approved):

thomasten (approved):

PR #174, opened 2021-06-02
make ttls clientAuth configurable

3u13r:
Previously whenever "Cert" was specified, we disabled ttls clientAuth.
Now the user can set clientAuth on incoming connections where a custom certificate is set. The client then needs to have a valid certificate signed by the coordinator's root CA. 
Also see https://github.com/edgelesssys/marblerun.sh/pull/49#discussion_r642926105

### DIFF
--- a/coordinator/core/marbleapi.go
+++ b/coordinator/core/marbleapi.go
@@ -357,7 +357,7 @@ func (c *Core) setTTLSConfig(marble manifest.Marble, specialSecrets reservedSecr
 
 				connConf["clicrt"] = stringUserClientCert
 				connConf["clikey"] = stringUserClientKey
-				connConf["clientAuth"] = false
+				connConf["clientAuth"] = !entry.DisableClientAuth
 				connConf["cacrt"] = stringCaCert
 
 			} else {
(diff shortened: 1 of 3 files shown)


thomasten:
(How) can this be used in practice?

3u13r:
I think that the question boils down to: do we need a second way (e.g. not with the marble's certificate) to wrap connections between marbles. For me this is also counterintuitive, but I could imagine that there are some niche configurations where this is wanted. 

Since I cannot think of any concrete example, I'm also fine with leaving the state as is. Maybe @m1ghtym0 has some more insights.

3u13r:
Added the changes from the discussion with @thomasten @m1ghtym0. Addionally added an error message when trying to set `DisableClientAuth=true` on incoming connections without an Cert defined. This reduces the use cases a bit more, since r.g. the web server in emojivoto could also use the marble's standard certificate. But I think with the corresponding documentation this enables a clearer distinction between marble-marble and marble-external connections.

thomasten (approved):

m1ghtym0 (approved):

PR #177, opened 2021-06-07
AB#999 Use cross-signed intermediate certificate

### LINKED ISSUE #175: Coordinator provisions intermediate certificate to marbles but not root CA certificate.
Hi Team,

Looking at this [line](https://github.com/edgelesssys/marblerun/blob/787f8458008197a41437bf38a3bb392ef95fa6e7/coordinator/core/marbleapi.go#L302) of code, it is understood that the coordinator simply provisions the intermediate CA certificate to marbles instead of the entire chain of certificate(specifically the Root CA certificate).

Using the provisioned Intermediate CA certificate to validate a leaf certificate leads to "unknow CA" alerts during TLS handshake.

Please help me with the following question.

**Can a client trust a server certificate simply using the Intermediate CA's certificate ? Or does it also need to have Root CA's (which signed the intermediate CA certificate ) certificate.**

I could face the above error while running the Redis server in a marble using [standalone](https://www.marblerun.sh/docs/deployment/standalone/)  mode. Then trying to establish TLS session with it using a sample Redis client running in a different marble.

Modification : Adding a note 
The marbles were authenticated and provisioned with secrets by the same coordinator

m1ghtym0:
Cross-sign the intermediate certificate.
The Marbles see a self-signed root certificate.
The "outside world" sees an intermediate certificate signed by the Coordinator's root CA.
Verifies the certificate chain with OpenSSL in the Marble API test.
Fixes #175

### DIFF
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -75,6 +75,15 @@ const coordinatorName string = "Marblerun Coordinator"
 // coordinatorIntermediateName is the name of the Coordinator. It is used as CN of the intermediate certificate which is set when setting or updating a certificate.
 const coordinatorIntermediateName string = "Marblerun Coordinator - Intermediate CA"
 
+// storage keys for the used in the Coordinator
+const (
+	sKCoordinatorRootCert         string = "coordinatorRootCert"
+	sKCoordinatorRootKey          string = "coordinatorRootKey"
+	skCoordinatorIntermediateCert string = "coordinatorIntermediateCert"
+	sKMarbleRootCert              string = "marbleRootCert"
+	sKCoordinatorIntermediateKey  string = "coordinatorIntermediateKey"
+)
+
 // Needs to be paired with `defer c.mux.Unlock()`
 func (c *Core) requireState(states ...state) error {
 	c.mux.Lock()
@@ -133,7 +142,7 @@ func NewCore(dnsNames []string, qv quote.Validator, qi quote.Issuer, sealer seal
 		if err := c.advanceState(stateRecovery); err != nil {
 			return nil, err
 		}
-	} else if _, err := c.store.getCertificate("root"); store.IsStoreValueUnsetError(err) {
+	} else if _, err := c.store.getCertificate(sKCoordinatorRootCert); store.IsStoreValueUnsetError(err) {
 		// no state was found, wait for manifest
 		c.zaplogger.Info("No sealed state found. Proceeding with new state.")
 		if err := c.setCAData(dnsNames); err != nil {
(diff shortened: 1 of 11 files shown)


coordinator/core/marbleapi_test.go
```diff
@@ -318,6 +326,37 @@ func (ms *marbleSpawner) verifyCertificateFromEnvironment(envName string, params
 	return *certificate
 }
 
+func (ms *marbleSpawner) verifyCertChainWithOpenSSL(pRoot, pLeaf, pKey *pem.Block) {
+	opensslCtx, err := openssl.NewCtx()
+	ms.require.NoError(err)
+	certStore := opensslCtx.GetCertificateStore()
+	opensslIntermediate, err := openssl.LoadCertificateFromPEM(pem.EncodeToMemory(pRoot))
+	ms.require.NoError(err)
+	openssLeaf, err := openssl.LoadCertificateFromPEM(pem.EncodeToMemory(pLeaf))
```
Nirusu, line 335: Typos down here.
Should bne: openss**l**Leaf/Key

coordinator/core/core.go
```diff
@@ -75,6 +75,13 @@ const coordinatorName string = "Marblerun Coordinator"
 // coordinatorIntermediateName is the name of the Coordinator. It is used as CN of the intermediate certificate which is set when setting or updating a certificate.
 const coordinatorIntermediateName string = "Marblerun Coordinator - Intermediate CA"
 
+// storage keys for the certificates used in the Coordinator
+const (
+	rootCAStore         string = "rootCA"
+	intermediateCAStore string = "intermediateCA"
+	marbleRootCAStore   string = "marbleRootCA"
```
thomasten, line 82:
```suggestion
	storeRootCA         = "rootCA"
	storeIntermediateCA = "intermediateCA"
	storeMarbleRootCA   = "marbleRootCA"
```
maybe prefix this to better group them, considering that this will grow?

coordinator/core/core.go
```diff
@@ -525,24 +535,36 @@ func (c *Core) generateCertificateForSecret(secret manifest.Secret, parentCertif
...
 		return err
 	}
-	if err := c.store.putPrivK("root", rootPrivK); err != nil {
+	if err := c.store.putPrivK(intermediateCAStore, intermediatePrivK); err != nil {
 		return err
 	}
-	if err := c.store.putPrivK("intermediate", intermediatePrivK); err != nil {
+	if err := c.store.putPrivK(marbleRootCAStore, intermediatePrivK); err != nil {
 		return err
 	}
```
thomasten, line 569:
```suggestion
```
I think we should store this only once (single source of truth). Maybe under a name that is neither "intermediate" nor "marbleRoot".

thomasten (approved):

PR #178, opened 2021-06-07
AB#997 add store transactions

daniel-weisse:
Introduces transactions to the store interface:
Every call to `Put` should only be done after a transaction has been started with `BeginTransaction`.
Once a transaction has successfully completed, the state can be sealed using `Commit`, if an error occurs the state is reverted to before `BeginTransactio` was called using `Rollback`.

Also fixes a problem where the coordinator state could not be sealed after recover when using the `NoEnclaveSealer` due to an unset encryption key.


thomasten:
I fixed some issues:
* Generalized the Store interface by removing Sealer artifacts
* There was a problem with the transactions as described in the ticket. Store.Get would see changes before they are committed. To fix this, I introduced a Transaction object.
* Changed the wrapper such that it can be used with either a Store or a Transaction
* Transactions must always be finished within an operation (startup, REST API call).

The transition to KV-Store is not completed yet. We will have to refactor some of the recovery as this only applies to StdStore, but not to other stores. (But not within this PR.)

m1ghtym0 (approved):

thomasten (approved):

Issue #179, opened 2021-06-08
Incomplete Coordinator's spec with Intel's SGX plugin


gianlu33:
Hello everyone,

**Describe the bug**

I tried to install Marblerun on a local Minikube cluster with Intel's SGX plugin, but the Coordinator is not able to generate a quote and thus the attestation fails.

Logs:
```
[meshentry] invoking premain
[meshentry] invoking main
{"level":"info","ts":1623160373.8714085,"caller":"coordinator/run.go:46","msg":"starting coordinator","version":"0.3.2","commit":"2afe8c096e4d7beab060983ef7ebcd1ad2445c3d"}
{"level":"info","ts":1623160373.8714085,"caller":"coordinator/run.go:56","msg":"creating the Core object"}
{"level":"info","ts":1623160373.8714085,"caller":"core/core.go:122","msg":"loading state"}
{"level":"info","ts":1623160373.8714085,"caller":"core/core.go:139","msg":"No sealed state found. Proceeding with new state."}
{"level":"info","ts":1623160373.8834083,"caller":"core/core.go:360","msg":"generating quote"}
[error_driver2api sgx_enclave_common.cpp:249] Enclave not authorized to run, .e.g. provisioning enclave hosted in app without access rights to /dev/sgx_provision. You need add the user id to group sgx_prv or run the app as root.
[load_pce ../pce_wrapper.cpp:175] Error, call sgx_create_enclave for PCE fail [load_pce], SGXError:4004.
{"level":"warn","ts":1623160373.8994083,"caller":"core/core.go:363","msg":"Failed to get quote. Proceeding in simulation mode."}
```

**To reproduce**
Just install Marblerun in a cluster configured with Intel's SGX plugin

**Expected behavior**
Since the Coordinator uses "in-proc" attestation to generate the quote, the container needs special privileges to launch enclaves with the provision bit set. Using the Intel's plugin, this can be achieved by specifing the following annotation in the metadata of the Pod spec:

```yaml
    metadata:
      annotations:
        sgx.intel.com/quote-provider: <name_of_container>
```
The result of this annotation is that the operator injects in the container's limits this line: `sgx.intel.com/provision: 1`
You can also see [in the Intel's official repository](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/deployments/sgx_enclave_apps/overlays/sgx_ecdsa_inproc_quote/kustomization.yaml) that the above annotation is used for this purpose.

[...]

PR #183, opened 2021-06-10
Fix CLI install for Intel device plugin

daniel-weisse:
This PR addresses an issue encountered when using the Intel SGX device plugin. (See https://github.com/edgelesssys/marblerun/issues/179#issue-915092193)
Since the coordinator uses in-process attestation we need to set `sgx.intel.com/provision` in the pods resource limits so the pod is allowed to access `/dev/sgx_provision`.

Additionally the EPC limits for the coordinator, as well as the injected EPC limits by the marble-injector, have been increased from 10 to 10Mi when using the Intel plugin.
This was an oversight when switching from the azure plugin which by default defined its limits in MiB.

[A PR for the helm chart](https://github.com/edgelesssys/helm/pull/10#issue-666031711) has also been created.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -416,6 +416,11 @@ func setSGXValues(resourceKey string, values, chartValues map[string]interface{}
 	if needNewLimit {
 		values["coordinator"].(map[string]interface{})["resources"].(map[string]interface{})["limits"].(map[string]interface{})[resourceKey] = 10
 	}
+
+	// Make sure provision bit is set if the Intel plugin is used
+	if resourceKey == intelEpc.String() {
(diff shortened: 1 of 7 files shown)


m1ghtym0:
This not going to age well. Can you keep the device plugin-related k8s information in one structure that implements functions for setting resource limits etc.? Whenever we add a new plugin or change smth. we only need to update this structure.

m1ghtym0:
I had a slightly different refactoring in mind, but we should do this in a different PR.
I suggest we wait if there's ever a different device plugin used by anyone and consider refactoring then.

m1ghtym0 (approved):

PR #184, opened 2021-06-10
AB#976 Add RBAC

daniel-weisse:
This PR changes how permissions work for users when interacting with the coordinator:

First up instead of `Admins` the manifest now defines `Users`. 
Users are ordered by their name and each define a `Certificate`, which is the same as before for `Admins`, as well as permissions to read/write secrets and update packages.
Functionality for these permissions will be added at a later date.

Functions related to users are in a new module `coordinator/user`.
From the user definition in a manifest, a `MarblerunUser` is created.
This type holds information about:
* the users name
* the users certificate
* the users permissions

The main difference for now is the change from `VerifyAdmin` to `VerifyUser` in the `ClientCore` interface. Which, instead of returning true or false, return the user matching the supplied certificate.

Using `marblerunUser.IsGranted(somePermission)` we can check if a user has been granted a requested permission.
This allows us to gate specific functions of the core to only privileged users.
For example allowing only user `A` and `B` to update a manifest, while the update by user `C` gets rejected.

### DIFF
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -570,7 +570,7 @@ func GenerateUsersFromManifest(rawUsers map[string]manifest.User) ([]*user.User,
 		newUser := user.NewUser(name, cert)
 		newUser.Assign(user.NewPermission(user.PermissionSetSecret, userData.WriteSecrets))
 		newUser.Assign(user.NewPermission(user.PermissionReadSecret, userData.ReadSecrets))
-		newUser.Assign(user.NewPermission(user.PermissionAllowedUpdate, userData.UpdatePackages))
+		newUser.Assign(user.NewPermission(user.PermissionUpdatePackage, userData.UpdatePackages))
 		users = append(users, newUser)
 	}
 	return users, nil
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -18,7 +18,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
-	"encoding/pem"
 	"errors"
 	"fmt"
 	"math"
@@ -39,6 +38,7 @@ import (
 
 // Core implements the core logic of the Coordinator
 type Core struct {
+	mux       sync.Mutex
 	quote     []byte
 	recovery  recovery.Recovery
 	store     store.Store
@@ -46,7 +46,6 @@ type Core struct {
 	sealer    seal.Sealer
 	qv        quote.Validator
 	qi        quote.Issuer
-	mux       sync.Mutex
 	zaplogger *zap.Logger
 }
 
@@ -60,14 +59,6 @@ const (
 	stateAcceptingMarbles
 	stateMax
 )
-
-// marblerunUser represents a privileged user of Marblerun
-type marblerunUser struct {
-	// name is the username
-	name string
-	// certificate is the users certificate, used for authentication
-	certificate *x509.Certificate
-}
 
 // coordinatorName is the name of the Coordinator. It is used as CN of the root certificate.
 const coordinatorName string = "Marblerun Coordinator"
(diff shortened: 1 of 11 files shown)


coordinator/user/user_test.go
```diff
@@ -0,0 +1,57 @@
...
+	ok = testUser.IsGranted(NewMarblerunPermission("unkownResource", []string{"perm-2"}))
+	assert.False(ok)
+}
+
+func TestGenerateUsersFromManifest(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+
+	var manifest manifest.Manifest
+	err := json.Unmarshal([]byte(test.ManifestJSONWithRecoveryKey), &manifest)
```
thomasten, line 51: I think it's better to have some specific test inputs here instead of using some manifest

coordinator/core/storewrapper.go
```diff
@@ -198,25 +204,38 @@ func (s storeWrapper) putState(currState state) error {
...
 }
 
 // putUser saves user information to store
-// will be changed in the future to set permissions etc. instead of just certificate
-func (s storeWrapper) putUser(user *marblerunUser) error {
-	request := strings.Join([]string{requestUser, user.name}, ":")
-	return s.store.Put(request, user.certificate.Raw)
+func (s storeWrapper) putUser(newUser *user.MarblerunUser) error {
+	request := strings.Join([]string{requestUser, newUser.Name()}, ":")
+	userData := &sealedUser{
```
thomasten, line 232: MarblerunUser should implement the json.Marshaler interface like we did here:
https://github.com/edgelesssys/marblerun/blob/1bf8e78988700886e453ff07f773af1d88068189/coordinator/manifest/manifest.go#L169-L197

thomasten (changes requested):
Thanks for this. Design looks quite clean.

coordinator/core/core_test.go
```diff
@@ -128,7 +128,7 @@ func TestRecover(t *testing.T) {
 	assert.Error(err)
 
 	// Set manifest. This will seal the state.
-	_, err = c.SetManifest(context.TODO(), []byte(test.ManifestJSON))
+	_, err = c.SetManifest(context.TODO(), []byte(test.ManifestJSONWithRecoveryKey))
```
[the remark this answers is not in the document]
daniel-weisse: I think the original intention was to test if user permissions persisted after sealing, but since this should be done in different places, this can be removed

coordinator/core/core.go
```diff
@@ -561,6 +554,28 @@ func (c *Core) generateCertificateForSecret(secret manifest.Secret, parentCertif
...
+		block, _ := pem.Decode([]byte(userData.Certificate))
+		if block == nil {
+			return nil, fmt.Errorf("received invalid certificate for user %s", name)
+		}
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return nil, err
+		}
+		newUser := user.NewUser(name, cert)
+		newUser.Assign(user.NewPermission(user.PermissionSetSecret, userData.WriteSecrets))
```
thomasten, line 571:
```suggestion
		newUser.Assign(user.NewPermission(user.PermissionWriteSecret, userData.WriteSecrets))
```
shouldn't this also be renamed then?

coordinator/core/core.go
```diff
@@ -561,6 +554,28 @@ func (c *Core) generateCertificateForSecret(secret manifest.Secret, parentCertif
 	return secret, nil
 }
 
+// GenerateUsersFromManifest creates users and permissions from a map of manifest.User
+func GenerateUsersFromManifest(rawUsers map[string]manifest.User) ([]*user.User, error) {
```
thomasten, line 558:
```suggestion
// generateUsersFromManifest creates users and permissions from a map of manifest.User
func generateUsersFromManifest(rawUsers map[string]manifest.User) ([]*user.User, error) {
```

coordinator/manifest/manifest.go
```diff
@@ -30,8 +30,8 @@ type Manifest struct {
 	Infrastructures map[string]quote.InfrastructureProperties
 	// Marbles contains the allowed services with their corresponding enclave and configuration parameters.
 	Marbles map[string]Marble
-	// Admins contains user-generated TLS client certificates to be used for an administrator to perform manifest updates
-	Admins map[string]string
+	// Users contains user-generated TLS client certificates to be used for an administrator to perform manifest updates
```
thomasten, line 33: something like
```suggestion
	// Users contains user definitions, including client certificates and permissions
```

thomasten (approved):

m1ghtym0 (approved):

PR #188, opened 2021-06-15
AB#988 Allow reading of secrets by users

daniel-weisse:
### Proposed changes
Allow users to read out secrets using the `/secrets` endpoint.
The manifest defines which secrets a user is allowed to read. The following shows an example of a user allowed to read the secrets `symmetric_key_shared` and `cert_shared`:
```javascript
{
//...
  "Users": {
    "marblerunUser": {
      "Certificate": "<certificate>",
      "ReadSecrets": [
        "symmetric_key_shared",
        "cert_shared" 
      ]
    }
  }
//...
}
```

### Additional info
- A user can not read private secrets (secrets without the `shared=true` property), because those are only accessible to the respective Marble
- The `/secrets` endpoint expects requests in the form of a query string in the following format:
```
https://<IP>:<PORT>/secrets?s=<secret_name_one>&s=<secret_name_two>&...
```
- Requests may specify one or more secrets, but will fail if the user is not allowed to read one of the secrets
- Requests where the query string contains empty fields for `s` will fail, e.g. `?s=key_shared&s=&s=cert_shared`
- Requests where the query string contains fields different than `s` will fail, e.g. `?s=key_shared&c=test` 
- Successful requests will contain the secrets as `map[string]manifest.Secret` marshaled to JSON in the `data` field of the JSON response

### DIFF
--- a/coordinator/server/server.go
+++ b/coordinator/server/server.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/edgelesssys/marblerun/coordinator/core"
 	"github.com/edgelesssys/marblerun/coordinator/rpc"
+	"github.com/edgelesssys/marblerun/coordinator/user"
 	"github.com/gorilla/handlers"
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_zap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
@@ -198,14 +199,8 @@ func CreateServeMux(cc core.ClientCore) *http.ServeMux {
 	})
 
 	mux.HandleFunc("/update", func(w http.ResponseWriter, r *http.Request) {
-		// Abort if no user client certificate was provided
-		if r.TLS == nil {
-			writeJSONError(w, "no client certificate provided", http.StatusUnauthorized)
-			return
-		}
-		user, err := cc.VerifyUser(r.Context(), r.TLS.PeerCertificates)
-		if err != nil {
-			writeJSONError(w, "unauthorized user", http.StatusUnauthorized)
+		user := verifyUser(w, r, cc)
+		if user == nil {
 			return
 		}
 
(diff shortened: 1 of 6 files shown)


m1ghtym0:
You're defining `ReadSecrets` as new property for users. Next, there will probably be `WriteSecrets` and `UpdatePackage`, etc.
Then we might want to add the same type of privileges to Marbles.
This will grow quite confusing soon.
What about having one central `Role` section in our manifest that stores roles for `Users` and `Marbles` (inspired by [k8s role bindings](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources)):

```
"Roles": {
   "cert-reader": {
        "resource": "Secrets",
        "resourceNames": ["symmetric_key_shared", "cert_shared"]
        "verbs": ["read"]
    }
},
"Users": {
    "marblerunUser": {
        "Certificate": "<certificate>",
        "Roles": ["cert-reader"]
    }
}
"Packages": {
    "frontend": {
        ...
        "Roles": ["cert-reader"]
    }
}
```

This would it make it very easy to comprehend what Users/Packages have access to what resources in Marblerun.
Any opinions @thomasten @Nirusu ?

Nirusu:
I think having Roles make sense as this will become more convoluted over time with more additions. Especially when we add this to Marbles, as then you start having duplicate definitions all over the place.

So I support the idea of introducing roles, though personally I would keep it simple and don't use K8s vocabulary (e.g. call it `actions`/`methods` instead of `verbs`). It's too implicit in my eyes.

daniel-weisse:
Roles sound like a good idea.

Since the changes necessary for this to happen basically come down to making a few adjustments to `coordinator/manifest/manifest.go` and the `generateUsersFromManifest` function from `coordinator/core/core.go`, should this maybe be done in a separate PR?
I could start a draft so we can discuss the proposed changes in more detail.

m1ghtym0:
> Roles sound like a good idea.
> 
> Since the changes necessary for this to happen basically come down to making a few adjustments to `coordinator/manifest/manifest.go` and the `generateUsersFromManifest` function from `coordinator/core/core.go`, should this maybe be done in a separate PR?
> I could start a draft so we can discuss the proposed changes in more detail.

Sure. Should this PR be on hold until then?

daniel-weisse:
Yes, lets wait until we have a proper role/permission format figured out and implemented.

coordinator/server/server.go
```diff
@@ -227,6 +227,46 @@ func CreateServeMux(cc core.ClientCore) *http.ServeMux {
...
+		// Abort if no user client certificate was provided
+		if r.TLS == nil {
+			writeJSONError(w, "no client certificate provided", http.StatusUnauthorized)
+			return
+		}
+		user, err := cc.VerifyUser(r.Context(), r.TLS.PeerCertificates)
+		if err != nil {
+			writeJSONError(w, "unauthorized user", http.StatusUnauthorized)
+			return
+		}
```
thomasten, line 240:
```suggestion
		user := verifyUser(w, r)
		if user == nil {
			return
		}
```

thomasten:
I think we should finish this PR first. The current permission format is not introduced in this PR, but it is already in master.

thomasten (approved):

m1ghtym0 (approved):

PR #189, opened 2021-06-15
Add Roles to manage Permissions

daniel-weisse:
As mentioned here: https://github.com/edgelesssys/marblerun/pull/188#issuecomment-861421271, having a centralised role system will make managing permissions easier when creating a manifest.

This PR serves to discuss the best format to do so.

For now I have created a sample manifest (`samples/sample-manifest.json`) to showcase a potential format.

For now a role defines 4 things:
* The name of the role
* The type of resource being affected (Secrets, Packages, etc.)
* The name of the resource (e.g. `symmetric_key_shared`)
* The actions allowed to perform on the resource (increment svn, read secret, write secret, etc.)

Things that remain a bit confusing and will need to be discussed:
* Not all resources allow the same actions: only `Packages` will allow incrementing, and for now only secrets allow reading the values
  * Should we use the same name for updating a package and setting a secret? e.g. the action `write` allows updates to packages and setting of secrets

### DIFF
--- a/coordinator/core/clientapi.go
+++ b/coordinator/core/clientapi.go
@@ -82,7 +82,7 @@ func (c *Core) SetManifest(ctx context.Context, rawManifest []byte) (map[string]
 	c.sealer.SetEncryptionKey(encryptionKey)
 
 	// Parse X.509 user certificates and permissions from manifest
-	users, err := generateUsersFromManifest(manifest.Users)
+	users, err := generateUsersFromManifest(manifest.Users, manifest.Roles)
 	if err != nil {
 		c.zaplogger.Error("Could not parse specified user certificate from supplied manifest", zap.Error(err))
 		return nil, err
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"strings"
 	"sync"
 	"time"
 
@@ -575,7 +576,7 @@ func generateUsersFromManifest(rawUsers map[string]manifest.User, roles map[stri
 		for _, assignedRole := range userData.Roles {
 			for _, action := range roles[assignedRole].Actions {
 				// correctness of roles has been verified by manifest.Check()
-				newUser.Assign(user.NewPermission(action, roles[assignedRole].ResourceNames))
+				newUser.Assign(user.NewPermission(strings.ToLower(action), roles[assignedRole].ResourceNames))
 			}
 		}
 		users = append(users, newUser)
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -559,7 +559,7 @@ func (c *Core) generateCertificateForSecret(secret manifest.Secret, parentCertif
 }
 
 // generateUsersFromManifest creates users and permissions from a map of manifest.User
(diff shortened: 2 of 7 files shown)


m1ghtym0:
This is mostly in line with my proposal, so I'm happy with this:-)
Actions naming is one thing we could discuss.
`read` and `write` are no-brainers. `update` needs to be handled very carefully. You've replaced it with `increment` to make it more explicit "what" can be updated. `increment` makes sense to us, but might be confusing for people less familiar with the concept. We could make it even more explicit with `UpdateSecurityVersion` or smth. or less explicit with just `update` and explanations in the docs.

thomasten:
We may support updating other parts of packages in future, so maybe better name it updateSecurityVersion than just update?

daniel-weisse:
Converting this to a PR since a working version is implemented.

For now a role looks as follows:
```javascript
"secret_role": {
  "ResourceType": "Secrets",
  "ResourceNames": [
    "secret_one",
    "secret_two"
  ],
  "Actions": [
    "write",
    "read"
  ]
},
"secret_role": {
  "ResourceType": "Packages",
  "ResourceNames": [
    "frontend",
    "backend"
  ],
  "Actions": [
    "updateSecurityVersion"
  ]
}
```

I'm still somewhat unsure about the naming of some things:

For now I chose the name `ResourceType`, which is a bit more verbose than the previously suggested `Resource`.
I think both options have their upsides and downsides, any opinions on this?

Currently all `Actions` are spelled lowercase, with actions on secrets being one verb, and actions on packages verb + object identifier.
Personally i think uppercase would look a bit nicer, but we could also just ignore spelling all together and accept both lower- and upper-case.
If we plan on implementing more features in the future which would allow the usage of `write` or `read` permissions, keeping the naming as is, is probably better. Otherwise a more verbose `ReadSecret` and `WriteSecret` might be preferable.

Nirusu:
I think switching to a more verbose way such as `ReadSecret` or `WriteSecret` might make more sense as then we always follow "verb + object", making everything look more consistent and not too different.

I think we can ignore case-sensitivity, we didn't care about it in other place too?

Otherwise one thing I could imagine is that having the "verb + object" verbose actions, we could even get rid of `ResourceType` and just let everything handle through the name, assuming duplicates are not possible in our Manifest. This would allow cross-resource-type roles, but I am not sure this is currently feasible (I guess duplicates are possible, as we store the objects in separate maps?), and I am also not sure if this a good idea in practice. 

But yeah, just food for thought I guess.

m1ghtym0:
> I think switching to a more verbose way such as ReadSecret or WriteSecret might make more sense as then we always follow "verb + object", making everything look more consistent and not too different.

Makes sense.

> Otherwise one thing I could imagine is that having the "verb + object" verbose actions, we could even get rid of ResourceType and just let everything handle through the name, assuming duplicates are not possible in our Manifest. This would allow cross-resource-type roles, but I am not sure this is currently feasible (I guess duplicates are possible, as we store the objects in separate maps?), and I am also not sure if this a good idea in practice.

I would refrain from cross-resource-type roles. Currently, duplicates are possible. Even without duplicates, this makes it harder to read and understand which resources correspond to which actions.

m1ghtym0 (approved):

PR #191, opened 2021-06-16
AB#987 Allow setting of secrets by users

daniel-weisse:
### Proposed changes
- Allow users to set secrets after the coordinator has been started and initialised with a manifest.

This is done using the `/secrets` endpoint of the client API and a special "secrets" manifest.
This manifest contains the data for one or more secrets marked as `UserDefined` in the original manifest and looks similar to the following:
```javascript
{
  "symmetric_key_unset": {
    "Key": "<base64 encoded key>"
  },
  "cert_unset": {
    "Cert": "<base64 encoded certificate>",
    "Private": "<optional base64 encoded private key of the cert>"
  },
  "generic_secret": {
    "Key": "<generic base64 encoded data>"
  }
}
```

In the original manifest `symmetric-key` and `cert` secrets are defined as before, apart from the extra `UserDefined` flag.
Plain secrets only require the type to be `plain` and the `UserDefined` flag.

A user trying to set a secret using the API has to have permissions for every secret they try to set.

A secret in the "secrets" manifest consist of 4 parts:
* The name of the secret
* `Key`: a base64 encoded symmetric key or arbitrary base64 encoded data
* `Cert`: a x509 certificate
* `Private`: a private key of a certificate, this can be left empty when uploading a certificate

When uploaded, the names of the secrets defined in the "secrets" manifest are matched against the ones defined in the original Manifest.
If no match is found, or the secret is not marked as `UserDefined` an error occurs, otherwise one of three cases happens:
* The secret is of type `symmetric-key`: 
  * the length of the data in `Key` is verified against the size defined in the original manifest
  * if `Key` or `Private` are set an error occurs
  * a new `manifest.Secret` of type `symmetric-key` is created using the metadata from the original manifest and key from the "secrets" manifest, and uploaded to the store
* The secret is of type `cert-*`:
  * if `Key` is set an error occurs
  * a new `manifest.Secret` of type `cert-*` is created using the metadata from the original manifest and cert, public and private key from the "secrets" manifest, and uploaded to the store
* The secret is of type `plain`:
  *  if `Key` or `Private` are set an error occurs
  * a new `manifest.Secret` of type `plain` is created using the metadata from the original manifest and key from the "secrets" manifest, and uploaded to the store

### Additional info
If a Marble requiring a secret marked as `UserDefined` tries to start before said secret is uploaded, the activation request fails.
The same thing happens when a Marble requiring the private key of a certificate marked as `UserDefined` tries to start, but the private key was not uploaded along the certificate.

### DIFF
--- a/coordinator/core/marbleapi.go
+++ b/coordinator/core/marbleapi.go
@@ -116,16 +116,8 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
 	}
 
 	// Union user-defined secrets with shared and unique secrets
-	userSecrets := findUserSecrets(marble.Parameters, mainManifest.Secrets)
-	for _, k := range userSecrets {
-		secret, err := c.data.getSecret(k)
-		if err != nil {
-			if store.IsStoreValueUnsetError(err) {
-				return nil, fmt.Errorf("secret %s needed but not set", k)
-			}
-			return nil, err
-		}
-		secrets[k] = secret
+	if err := c.addUserSecrets(secrets, marble.Parameters, mainManifest.Secrets); err != nil {
+		return nil, err
 	}
 
 	// add TTLS config to Env
@@ -344,6 +336,22 @@ func parseSecrets(data string, secretsWrapped secretsWrapper) (string, error) {
 	return templateResult.String(), nil
 }
 
+func (c *Core) addUserSecrets(secrets map[string]manifest.Secret, params *rpc.Parameters, sharedSecrets map[string]manifest.Secret) error {
+	userSecrets := findUserSecrets(params, sharedSecrets)
+	for _, k := range userSecrets {
+		secret, err := c.data.getSecret(k)
+		if err != nil {
+			if store.IsStoreValueUnsetError(err) {
+				return fmt.Errorf("secret %s needed but not set", k)
+			}
+			return err
+		}
+		secrets[k] = secret
+	}
+
+	return nil
+}
+
 func (c *Core) generateMarbleAuthSecrets(req *rpc.ActivationReq, marbleUUID uuid.UUID) (reservedSecrets, error) {
 	// generate key-pair for marble
 	privk, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
@@ -483,7 +491,7 @@ func findUserSecrets(params *rpc.Parameters, secrets map[string]manifest.Secret)
 	// 1: the encoding
 	// 2: the name of the secret
 	// 3: if specified the part of the secret, empty string otherwise
-	r := regexp.MustCompile("{{\\s*(raw|hex|base64|pem)\\s+\\.Secrets\\.([\\w-]+)(\\.[\\w-]+)?\\s*}}")
+	r := regexp.MustCompile(`{{\s*(raw|hex|base64|pem)\s+\.Secrets\.([\w-]+)(\.[\w-]+)?\s*}}`)
 	wantedSecrets := make(map[string]bool)
 	for _, v := range params.Argv {
 		subStrings := r.FindAllStringSubmatch(v, -1)
--- a/coordinator/core/marbleapi.go
+++ b/coordinator/core/marbleapi.go
@@ -15,7 +15,9 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
+	"fmt"
 	"math"
+	"regexp"
 	"text/template"
 	"time"
 
@@ -92,14 +94,14 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
 		return nil, err
 	}
 
-	// Generate user-defined unique (= per marble) secrets
+	// Generate unique (= per marble) secrets
 	secrets, err := c.generateSecrets(ctx, mainManifest.Secrets, marbleUUID, marbleRootCert, intermediatePrivK)
 	if err != nil {
 		c.zaplogger.Error("Could not generate specified secrets for the given manifest.", zap.Error(err))
 		return nil, err
 	}
 
-	// Union user-defined unique secrets with user-defined shared secrets
+	// Union unique secrets with shared secrets
 	sharedSecrets, err := c.data.getSecretMap()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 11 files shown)


test/manifests.go
```diff
@@ -407,6 +455,17 @@ const UpdateManifest = `{
 	}
 }`
 
+// SecretsManifest is a test secrets manifest to update secrets
+const SecretsManifest = `{
```
thomasten, line 459: Maybe we shouldn't name all json documents manifest. We might only name things "SomethingManifest" if they keep the same format as the Manifest like the UpdateManifest does. Opinions? @m1ghtym0 @Nirusu

coordinator/core/marbleapi.go
```diff
@@ -112,6 +114,20 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
...
+	for _, k := range userSecrets {
+		secret, err := c.data.getSecret(k)
+		if err != nil {
+			if store.IsStoreValueUnsetError(err) {
+				return nil, fmt.Errorf("secret %s needed but not set", k)
+			}
+			return nil, err
+		}
+		secrets[k] = secret
+	}
```
thomasten, line 129:
```suggestion
	if err := c.addUserSecrets(secrets, marble.Parameters, mainManifest.Secrets); err != nil {
		return err
	}
```
we should keep the Activate func rather high level, delegating tasks to subfunctions

coordinator/core/marbleapi.go
```diff
@@ -458,3 +474,42 @@ func (c *Core) setTTLSConfig(marble manifest.Marble, specialSecrets reservedSecr
...
+
+// findUserSecrets finds all user-defined secrets required by a Marble
+func findUserSecrets(params *rpc.Parameters, secrets map[string]manifest.Secret) []string {
+	// Find the name of all secrets required by the marble
+	// For each match in a Argv/Env/File the sub-strings will be as follows:
+	// 0: the complete secret injection string ({{ <encoding> .Secrets.<name>.<part> }})
+	// 1: the encoding
+	// 2: the name of the secret
+	// 3: if specified the part of the secret, empty string otherwise
+	r := regexp.MustCompile("{{\\s*(raw|hex|base64|pem)\\s+\\.Secrets\\.([\\w-]+)(\\.[\\w-]+)?\\s*}}")
```
thomasten, line 486:
```suggestion
	r := regexp.MustCompile(`{{\s*(raw|hex|base64|pem)\s+\.Secrets\.([\w-]+)(\.[\w-]+)?\s*}}`)
```

thomasten, line 486: Any chance to use something from the `template` package to do this job? @m1ghtym0 @Nirusu

test/integration_test.go
```diff
@@ -274,6 +275,30 @@ func TestClientAPI(t *testing.T) {
 	resp.Body.Close()
 	require.NoError(err)
 	assert.Contains(string(secret), `{"status":"success","data":{"symmetric_key_shared":{"Type":"symmetric-key","Size":128,`)
+
+	// start server
```
thomasten, line 279: this should be a new test

coordinator/core/marbleapi.go
```diff
@@ -112,6 +114,20 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
 	if marble.Parameters == nil {
 		marble.Parameters = &rpc.Parameters{}
 	}
+
+	// Union user-defined secrets with shared and unique secrets
+	userSecrets := findUserSecrets(marble.Parameters, mainManifest.Secrets)
```
thomasten, line 119: add test in marbleapi_test.go

thomasten (changes requested):

test/manifests.go
m1ghtym0: Yep, agreed.

daniel-weisse: Originally this data structure was directly marshaled into a `manifest.Manifest` object. 
I agree that at the current state it would probably be better renaming it to something simpler like `UserSecrets`

coordinator/core/marbleapi.go
daniel-weisse: Regarding extracting the actions from a parsed template, there is this stackoverflow discussion: https://stackoverflow.com/questions/40584612/how-to-get-a-map-or-list-of-template-actions-from-a-parsed-template

If we were to implement it this way we would have to extract the actions from the parsed template for each Arg/Env/File, then check for only the actions requesting a value from `.Secrets`. The later step most likely being done with regex, at which point we might as well just pass over the entire string with regex and save us the template parsing

thomasten: Can we revert the changes in marbleapi.go and storewrapper.go and just don't accept empty values in EncodeSecretDataToRaw and EncodeSecretDataToPem?

daniel-weisse: I made a pass over the template functions., they now take care of returning errors when requested secret is not set.
This removes most of the changes to `marbleapi.go`.
`getSecretMap` from `storewrapper.go` now returns a map of all non unique secrets, including user defined ones.
If a secret is not yet set, the function will return an empty secret for that key.

coordinator/manifest/manifest.go
```diff
@@ -235,13 +235,22 @@ func (c *Certificate) UnmarshalJSON(data []byte) error {
 func EncodeSecretDataToPem(data interface{}) (string, error) {
```
thomasten, line 235:
```go
func EncodeSecretDataToPem(data interface{}) (string, error) {
	var typ string
	var bytes []byte
	switch secret := data.(type) {
	case Certificate:
		typ, bytes = "CERTIFICATE", secret.Raw
	case PublicKey:
		typ, bytes = "PUBLIC KEY", secret
	case PrivateKey:
		typ, bytes = "PRIVATE KEY", secret
	default:
		return "", errors.New("invalid secret type")
	}
	if len(bytes) <= 0 {
		return "", errors.New("tried to parse secret with empty value")
	}
	return string(pem.EncodeToMemory(&pem.Block{Type: typ, Bytes: bytes})), nil
}
```

coordinator/manifest/manifest.go
```diff
@@ -262,14 +271,29 @@ func EncodeSecretDataToHex(data interface{}) (string, error) {
 func EncodeSecretDataToRaw(data interface{}) (string, error) {
```
thomasten, line 271:
```go
func EncodeSecretDataToRaw(data interface{}) (string, error) {
	var result []byte
	switch secret := data.(type) {
	case []byte:
		result = secret
	case PrivateKey:
		result = secret
	case PublicKey:
		result = secret
	case Secret:
		result = secret.Public
	case Certificate:
		result = secret.Raw
	default:
		return "", errors.New("invalid secret type")
	}
	if len(result) <= 0 {
		return "", errors.New("tried to parse secret with empty value")
	}
	return string(result), nil
}
```

test/integration_test.go
```diff
@@ -276,6 +276,56 @@ func TestClientAPI(t *testing.T) {
...
+	// start a marble
+	log.Println("Starting a Client-Marble with unset secret, this should fail...")
+	clientCfg := newMarbleConfig(meshServerAddr, "test_marble_unset", "client,frontend,localhost")
+	defer clientCfg.cleanup()
+	assert.False(startMarbleClient(clientCfg))
+
+	// test setting a secret
+	log.Println("Setting a custom secret")
+	clientAPIURL := url.URL{Scheme: "https", Host: clientServerAddr, Path: "secrets"}
+	_, err = client.Post(clientAPIURL.String(), "application/json", bytes.NewReader([]byte(UserSecrets)))
```
thomasten, line 321: what's the difference between this and `strings.NewReader(UserSecrets)`?

daniel-weisse: No real difference, i saw `bytes.NewReader` being used everywhere else in the integration test and switched it.

thomasten: AFAIK it's best to use the one that doesn't require a cast

daniel-weisse: Makes sense, I'll look over the code and see where the `strings` function can be used over `bytes`

thomasten (approved):

PR #192, opened 2021-06-17
Fix/graphene samples

3u13r:
### Proposed changes
- graphene-hello: update manifest & makefile
- graphene-nginx: update manifest & makefile
- graphene-redis: update manifest, build graphene in docker 
- all of the above now works with the latest graphene master (https://github.com/oscarlab/graphene/tree/b37ac75efec0c1183fd42340ce2d3e04dcfb3388)

### Additional info
- currently broken down into 3 commits: hopefully helps reviewing 
- will squash into one commit after approval
- Note: our current redis-graphene-marble image needs to be updated

### DIFF
--- a/samples/graphene-hello/hello.manifest.template
+++ b/samples/graphene-hello/hello.manifest.template
@@ -1,8 +1,8 @@
-loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
+loader.preload = "file:{{ graphene.libos }}"
 loader.env.LD_LIBRARY_PATH = "/lib"
 
 # entrypoint must be premain-graphene
-libos.entrypoint = "file:premain-graphene"
+libos.entrypoint = "premain-graphene"
 sgx.trusted_files.premain = "file:premain-graphene"
 
 # argv0 must be the path to the actual application
@@ -13,11 +13,11 @@ loader.insecure__use_host_env = 1
 
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
-fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
+fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.ld = "file:$(GRAPHENEDIR)/Runtime/ld-linux-x86-64.so.2"
-sgx.trusted_files.libc = "file:$(GRAPHENEDIR)/Runtime/libc.so.6"
-sgx.trusted_files.pthread = "file:$(GRAPHENEDIR)/Runtime/libpthread.so.0"
+sgx.trusted_files.ld = "file:{{ graphene.runtimedir() }}/ld-linux-x86-64.so.2"
+sgx.trusted_files.libc = "file:{{ graphene.runtimedir() }}/libc.so.6"
+sgx.trusted_files.pthread = "file:{{ graphene.runtimedir() }}/libpthread.so.0"
 sgx.trusted_files.hello = "file:hello"
 
 # allow the marble's uuid file
@@ -34,7 +34,11 @@ sgx.thread_num = 16
 fs.mount.etc.type = "chroot"
 fs.mount.etc.path = "/etc"
 fs.mount.etc.uri = "file:/etc"
-sgx.trusted_files.nss = "file:$(GRAPHENEDIR)/Runtime/libnss_dns.so.2"
-sgx.trusted_files.resolv = "file:$(GRAPHENEDIR)/Runtime/libresolv.so.2"
+sgx.trusted_files.nss = "file:{{ graphene.runtimedir() }}/libnss_dns.so.2"
+sgx.trusted_files.resolv = "file:{{ graphene.runtimedir() }}/libresolv.so.2"
 sgx.allowed_files.hosts = "file:/etc/hosts"
 sgx.allowed_files.resolv = "file:/etc/resolv.conf"
+
+
+# Name Service Switch (NSS) files, see 'man nsswitch.conf'
+sgx.allowed_files.nsswitch  = "file:/etc/nsswitch.conf"
--- a/samples/graphene-redis/Dockerfile
+++ b/samples/graphene-redis/Dockerfile
@@ -9,18 +9,33 @@ WORKDIR /premain/build
 RUN cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
 RUN make premain-graphene
 
-# Use with fully built graphene as build context
-# Place redis-server.manifest.template from this repo inside $GRAPHENE_DIR/Examples/redis/
 FROM ghcr.io/edgelesssys/edgelessrt-deploy:latest AS release
-RUN apt-get update && apt-get install -y build-essential autoconf gawk bison wget python3 libcurl4-openssl-dev \
-    python3-protobuf libprotobuf-c-dev protobuf-c-compiler python3-pip software-properties-common
+RUN apt-get update && apt-get install -y git meson build-essential autoconf gawk bison wget python3 libcurl4-openssl-dev \
+    python3-protobuf libprotobuf-c-dev protobuf-c-compiler python3-pip software-properties-common python3-click python3-jinja2 \
+    linux-headers-`uname -r`
 RUN wget -qO- https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | apt-key add
 RUN add-apt-repository 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu bionic main'
 RUN apt-get install -y libsgx-quote-ex-dev libsgx-aesm-launch-plugin
-RUN python3 -m pip install toml
-COPY . /graphene
-COPY --from=build-premain /premain/build/premain-graphene /graphene/Examples/redis
+RUN python3 -m pip install "toml>=0.10"
+
+RUN git clone https://github.com/intel/SGXDataCenterAttestationPrimitives.git /SGXDriver
+WORKDIR /SGXDriver
+RUN git reset --hard a93785f7d66527aa3bd331ba77b7993f3f9c729b
+
+
+
+RUN git clone https://github.com/oscarlab/graphene.git /graphene
+WORKDIR /graphene
+RUN git reset --hard b37ac75efec0c1183fd42340ce2d3e04dcfb3388
+RUN make ISGX_DRIVER_PATH=/SGXDriver/driver/linux/ SGX=1
+RUN meson build -Ddirect=disabled -Dsgx=enabled
+RUN ninja -C build
+RUN ninja -C build install
+
+COPY --from=build-premain /premain/build/premain-graphene /graphene/Examples/redis/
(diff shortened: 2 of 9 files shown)


m1ghtym0 (approved):
Good job! Thanks!

samples/graphene-redis/Dockerfile
```diff
@@ -12,15 +12,26 @@ RUN make premain-graphene
...
+RUN apt-get update && apt-get install -y git meson build-essential autoconf gawk bison wget python3 libcurl4-openssl-dev \
+    python3-protobuf libprotobuf-c-dev protobuf-c-compiler python3-pip software-properties-common python3-click python3-jinja2 \
+    linux-headers-$(uname -r)
 RUN wget -qO- https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | apt-key add
 RUN add-apt-repository 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu bionic main'
 RUN apt-get install -y libsgx-quote-ex-dev libsgx-aesm-launch-plugin
-RUN python3 -m pip install toml
-COPY . /graphene
-COPY --from=build-premain /premain/build/premain-graphene /graphene/Examples/redis
+RUN python3 -m pip install toml>=0.10
```
daniel-weisse, line 21:
```suggestion
RUN python3 -m pip install "toml>=0.10"
```
running this without quotation marks leads to installation of `toml`, not necessarily at version `>=0.10`, with the output piped to a file called `=0.10`

samples/graphene-hello/hello.manifest.template
```diff
@@ -34,7 +34,11 @@ sgx.thread_num = 16
...
-sgx.trusted_files.nss = "file:$(GRAPHENEDIR)/Runtime/libnss_dns.so.2"
-sgx.trusted_files.resolv = "file:$(GRAPHENEDIR)/Runtime/libresolv.so.2"
+sgx.trusted_files.nss = "file:{{ graphene.runtimedir() }}/libnss_dns.so.2"
+sgx.trusted_files.resolv = "file:{{ graphene.runtimedir() }}/libresolv.so.2"
 sgx.allowed_files.hosts = "file:/etc/hosts"
 sgx.allowed_files.resolv = "file:/etc/resolv.conf"
+
+
+# Name Service Switch (NSS) files, see 'man nsswitch.conf'
+sgx.allowed_files.nsswitch  = "file:/etc/nsswitch.conf"
```
thomasten, line 44:
```suggestion
sgx.allowed_files.nsswitch  = "file:/etc/nsswitch.conf"
```
Do we need nsswitch at all?

samples/graphene-redis/redis-manifest.json
```diff
@@ -1,10 +1,7 @@
 {
     "Packages": {
         "redis-package": {
-            "Debug": true,
-            "SignerID": "43361affedeb75affee9baec7e054a5e14883213e5a121b67d74a0e12e9d2b7a",
-            "ProductID": 13,
-            "SecurityVersion": 1
+            "Debug": true
```
thomasten, line 4:
```suggestion
            "Debug": true,
            "SignerID": "43361affedeb75affee9baec7e054a5e14883213e5a121b67d74a0e12e9d2b7a",
            "ProductID": 13,
            "SecurityVersion": 1
 ```

File: samples/graphene-redis/redis-server.manifest.template [hunk 4]
```diff
@@ -162,15 +162,14 @@ sgx.allowed_files.ethers = "file:/etc/ethers"
...
+# MARBLERUN: allow aesm socket
 sgx.allowed_files.aesm_socket = "file:/var/run/aesmd/aesm.socket"
 
 # getaddrinfo(3) configuration file. Glibc reads this file to correctly find
 # network addresses. For more info, see 'man gai.conf'.
 sgx.allowed_files.gaiconf = "file:/etc/gai.conf"
-
-############################# SGX: Attestation ###############################
-sgx.isvprodid = 13
-sgx.isvsvn    = 1
```
thomasten, line 176: revert this?

thomasten (changes requested):

samples/graphene-hello/hello.manifest.template
3u13r: I added this, because the application tried to access it, complained that it had no access and paniced. Was fixed by adding this line. Maybe there is a way, so that the application doesn't try to access it?

samples/graphene-redis/redis-manifest.json
3u13r: I changed this, because I had issues with the activation of the marble. From the documentation I thought that if `debug` is set, then the other values don't matter. This seems not to be the case. 
> set to true if the enclave is to be run in debug mode. This allows you to experiment with deploying your application with Marblerun without having to worry about setting correct values for the above properties, but note that enclaves in debug mode are not secure.
I had an issue there: https://github.com/edgelesssys/marblerun/blob/787f8458008197a41437bf38a3bb392ef95fa6e7/coordinator/quote/ert.go#L44
Maybe I just had issue because I removed the last lines in the graphene manifest because there was no MARBLERUN comment preceeding them. Will check that on tomorrow.

samples/graphene-hello/hello.manifest.template
thomasten: It's ok to add if it's needed. Just wondering why it's needed now and wasn't before.

samples/graphene-redis/redis-manifest.json
thomasten: If debug is true, the other values can be omitted. If it is false, the other values must be defined.
If they are defined, they are always checked, regardless of whether debug is true or false.
You had issues because you did not sign with the private key that has this signerid.

Nirusu: Why do we hardcode some value for SignerID there if it's not embedded in the Dockerfile? I am confused.

samples/graphene-redis/Dockerfile
```diff
@@ -12,15 +12,26 @@ RUN make premain-graphene
...
+RUN ninja -C build
+RUN ninja -C build install
+
+COPY --from=build-premain /premain/build/premain-graphene /graphene/Examples/redis/
+COPY redis-server.manifest.template /graphene/Examples/redis/
 WORKDIR /graphene/Examples/redis
 ENV BUILD_TLS yes
 RUN make clean && make SGX=1
-ENTRYPOINT [ "bash", "-c", "SGX=1 /graphene/Runtime/pal_loader /graphene/Examples/redis/redis-server" ]
+ENTRYPOINT [ "bash", "-c", "graphene-sgx /graphene/Examples/redis/redis-server" ]
```
daniel-weisse, line 37:
```suggestion
ENTRYPOINT [ "graphene-sgx" "/graphene/Examples/redis/redis-server" ]
```
I would assume this to work now, unless `graphene-sgx` is only added to PATH through `.bashrc` or something similar

Nirusu:
```
$ whereis graphene-sgx
graphene-sgx: /usr/local/bin/graphene-sgx
```

Should be in the default PATH, but otherwise we can also expect it to be in /usr/local/bin

samples/graphene-redis/README.md
```diff
@@ -108,18 +108,14 @@ You can now securely connect to the Redis server using the `redis-cli` and the M
...
 
-*Prerequisite*: Graphene is set up and the original Redis example is working.
-
 To marbleize the example we edited [redis-server.manifest.template](redis-server.manifest.template). See comments starting with `MARBLERUN` for explanations of the required changes.
 
-Assuming you have built Graphene in `/graphene` copy the redis-server.manifest.template into the `/graphene/Examples/redis`
-
 ```bash
 cp ./redis-server.manifest.template /graphene/Examples/redis/redis-server.manifest.template
 ```
```
daniel-weisse, line 115:
```suggestion
```
Should not be needed anymore since the file is copied during `docker build` from the current directory

daniel-weisse (approved):

samples/graphene-redis/Dockerfile
```diff
@@ -12,15 +12,26 @@ RUN make premain-graphene
...
 RUN apt-get install -y libsgx-quote-ex-dev libsgx-aesm-launch-plugin
-RUN python3 -m pip install toml
-COPY . /graphene
-COPY --from=build-premain /premain/build/premain-graphene /graphene/Examples/redis
+RUN python3 -m pip install "toml>=0.10"
+
+RUN git clone https://github.com/oscarlab/graphene.git /graphene
+WORKDIR /graphene
+RUN git reset --hard b37ac75efec0c1183fd42340ce2d3e04dcfb3388
+RUN make ISGX_DRIVER_PATH=/usr/src/linux-headers-5.4.0-1049-azure/arch/x86/ SGX=1
```
thomasten, line 26: Is this ok @m1ghtym0 @daniel-weisse ?

daniel-weisse: I'm not familiar with the best practice for this case but `git checkout b37ac75efec0c1183fd42340ce2d3e04dcfb3388` would also be an option

thomasten: Sorry, my comment was about line 26

3u13r: Just to explain the idea: I pinned the headers instead of doing `uname -r`. Those headers are needed for the AKS deployment which in my mind is the only way of deploying the container anyway. Furthermore, this way I am able to build the container outside of an SGX machine. I think another option would be to demand building the image on the same machine (e.g.  DC2s_v2) as they run on. 
Is this what you asked about?

Nirusu: Isn't this gonna break badly upon a kernel update (which are common)?

3u13r: That's a good point. Since I don't know what their update policies are for their confidential aks cluster are, we should revert this back to `uname -r` ? Personally, I would much rather look up the headers once on an ACCVM and then be good to the next upgrade then to have to build the images on ACCVM itself. Is there maybe another solution? For example, remotely querying the current version?

thomasten (approved):

samples/graphene-redis/Dockerfile
Nirusu: Can't we use the path to the SGX driver sources directly? Or ISGX_DRIVER_PATH=""? Though I am not sure the latter one works. It says it should use the in-kernel driver sources, but I am not sure if this means that only works with Linux 5.11 with the upstreamed driver or previous one. 

Nethertheless, we could specify the source directory of the isgx driver of the kernel module installed. We could get this from the list of loaded kernel modules, download it somewhere and then use this as a source directory?

m1ghtym0: @3u13r So we're pinning it to the current Azure kernel now?
If so could you add a short disclaimer to the READMEs?

3u13r: I wanted to do that, but then @Nirusu came up with other ideas. If no one has a clear cut solution for this, I'll probably add the disclaimer, and merge this. Then either graphene changes their build or we can implement more sophisticated ideas.

Nirusu: Don't like the idea of pinning this to a specific version, sorry.

PR #195, opened 2021-06-22
AB#1022 Unify libOS premains

Nirusu:
### Proposed changes
- Since Occlum 0.23.0 now support `execve` and fixed our DNS woes when built with our toolchain, we could merge both premains into one.
- This PR gets rid of `premain-graphene` & `premain-occlum` and introduces a common one which uses an `uname`-based detection under which libOS we are currently running. These are based on constants in both codebases, so if they ever change in the future, this might break (though so far they did not change for a couple of years).

### Related issue
- **Merge either #194 or this one.**  This depends on whether we decide to support two premains (#194) or if we want to unify them (this PR). 
- Will also properly break #192.

### Additional info
- `make` does not work so far automatically due to it trying to download `libos-premain` from GitHub. Of course this will only work when we add it to our releases. This can be skipped or replaces with a local download for testing.


thomasten:
I don't have a strong opinion yet. Having only one binary is probably good for UX. On the other hand, I have concern about the detection logic. And it may happen that in future we may integrate different runtimes and libos's with different methods.
Opinions @m1ghtym0 @daniel-weisse ?

daniel-weisse:
For usability having a single binary is probably the best. Not needing to worry about having to download the correct binary may cut down on potential user error.
If we have a sound detection logic, i don't see much reason to keep multiple premain binaries.
Ideally we could get LibOS developers to agree on a common way to signal their use. Looking at Occlum it seems rather straight forward to detect, if others work similarly detection shouldn't be a problem

m1ghtym0:
Definitely like the fact that we only have one binary. I only see Mystikos as on the horizon of future LibOSes and I'm with Nils here that they'll all behave fairly similar detection-wise. Regarding future runtimes, the main thing for me is Rust which is probably not a concern for the libOS premain.

daniel-weisse (approved):

Nirusu:
Fun fact: Mystikos wants to follow a certain pattern for `uname`:
https://github.com/deislabs/mystikos/blob/8cf3b9427d83c5e5296579389c6d22c87580d769/kernel/syscall.c#L1941

```c
long myst_syscall_uname(struct utsname* buf)
{
    // We are emulating Linux syscalls. 5.4.0 is the LTS release we
    // try to emulate. The release number should be updated when
    // Mystikos adapts to syscall API changes in future Linux releases.
    MYST_STRLCPY(buf->sysname, "Linux");
    MYST_STRLCPY(buf->release, "5.4.0");
    MYST_STRLCPY(buf->version, "Mystikos 1.0.0");
    MYST_STRLCPY(buf->machine, "x86_64");

    myst_spin_lock(&_hostname_lock);
    MYST_STRLCPY(buf->nodename, _hostname);
    myst_spin_unlock(&_hostname_lock);

    return 0;
}
```

Which is a bit interesting, as Graphene comments that they hardcoded uname because it's a fairly uninteresting syscall.

So far only Graphene seems to be "hiding" its own name to `uname`. But I think this seems doable for the time being, as we are not aware of any other libOSes, and these projects usually do not gain traction very fast due to their complexity. And if a new one appears, we could always reevaluate our decision here.

m1ghtym0 (approved):
Thanks!
Please adjust the docs accordingly:-)

Nirusu:
Updated in: https://github.com/edgelesssys/marblerun.sh/commit/7d1c55dfe320f2c38558c876df68db203e978dc1

Nirusu:
Rebased with changes from #192

m1ghtym0:
LGTM!

PR #196, opened 2021-06-24
AB#998 Replace update manifest with direct updates

daniel-weisse:
### Proposed changes
Directly update the manifest instead of keeping an extra updateManifest around.

The current structure of the core allows us to simply increase the security version of the main manifest and then save it to the store.
This saves us some extra checks where the core had to not only verify the security version of the main, but also the updateManifest.


Nirusu:
I am confused. I thought the old idea we had was to not change the SHA256 hash of the originally deployed manifest whenever we "just" bump the SecurityVersion. That was the reason why we had two different manifests, at least I thought that was why.

Do we not care about this anymore?

daniel-weisse:
You are right, this is something i didn't think of.
We could keep the hash of the original manifest around and serve it on request.
Or alternatively change the current PR to merge a new update manifest into an old one. That way we keep the distinction between main manifest and update, but repeated updates don't require specifying all already updated packages. (Maybe misunderstood the ticket and the later case was the actual idea)

daniel-weisse:
Closing for now, this will get reworked and implemented in a different way at a later date.

PR #197, opened 2021-06-25
AB#1006 collect Prometheus metrics

katexochen:
### Proposed changes
Collect and expose the following Prometheus metrics:
- Marble activation (activation attempts/ success), labeled with the marble's type and UUID
- Current state of the Coordinator
- gRPC server metrics of the Marble API
- HTTP server metrics of the client API
- Info metrics on the Coordinators version and commit

Further changes:
- Custom Prometheus registry to eliminate global state
- Use of Prometheus' promauto package to ensure registration of collectors

### DIFF
--- a/cmd/coordinator/run.go
+++ b/cmd/coordinator/run.go
@@ -18,6 +18,8 @@ import (
 	"github.com/edgelesssys/marblerun/coordinator/seal"
 	"github.com/edgelesssys/marblerun/coordinator/server"
 	"github.com/edgelesssys/marblerun/util"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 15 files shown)


coordinator/core/storewrapper.go
```diff
@@ -34,6 +34,7 @@ type storeWrapper struct {
 		Get(string) ([]byte, error)
 		Put(string, []byte) error
 	}
+	metrics *StoreWrapperMetrics // can be nil
```
thomasten, line 37:
```suggestion
```
storeWrapper is supposed to be completely stateless. I think callers of putState should update the metric. There are actually only two relevant callers.

thomasten (changes requested):

cmd/coordinator/run.go
```diff
@@ -53,24 +55,37 @@ func run(validator quote.Validator, issuer quote.Issuer, sealDir string, sealer
 	meshServerAddr := util.Getenv(config.MeshAddr, config.MeshAddrDefault)
 	promServerAddr := os.Getenv(config.PromAddr)
 
+	// Create Prometheus resources and start the Prometheus server.
+	var promRegistry *prometheus.Registry
+	var promFactory promauto.Factory
+	if promServerAddr != "" {
```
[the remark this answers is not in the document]
m1ghtym0: We don't consider the Prometheus logging security-relevant because all information logged can essentially be obtained by observing the cluster as an untrusted party.
So for the sake of simplicity, I'd keep it out of the manifest.

m1ghtym0:
I'm not sure why everything is on "resolved" but there is no commit that solves them, maybe a missing push?
@katexochen If you fix the last thing Thomas mentioned we can merge this:-)

katexochen:
Sorry, I didn't manage to finish this in time on Thursday. I had to additionally fix a bug where the state of the Coordinator wouldn't be collected correctly at recovery. I'm still not fully satisfied with the "state" metric collection, I think letting the callers of putState increment it has a great potential for people to forget. Also, it is really hard to test this metric. Maybe we should keep an eye on this and try to "grow" the collection of metrics more into the Coordinators design over time...

coordinator/core/core.go
```diff
@@ -136,6 +143,9 @@ func NewCore(dnsNames []string, qv quote.Validator, qi quote.Issuer, sealer seal
 			if err := txdata.putState(stateUninitialized); err != nil {
 				return nil, err
 			}
+			if c.metrics != nil {
+				c.metrics.coordinatorState.Set(float64(stateUninitialized))
+			}
```
thomasten, line 148:
```suggestion
```
Not needed. The function cannot succeed with this state.

coordinator/core/core.go
```diff
@@ -162,6 +172,9 @@ func NewCore(dnsNames []string, qv quote.Validator, qi quote.Issuer, sealer seal
 		if err := txdata.putState(stateAcceptingManifest); err != nil {
 			return nil, err
 		}
+		if c.metrics != nil {
```
thomasten, line 175: We should get rid of all these nil checks and use the null object pattern instead. To this end, you could implement the prometheus.Gauge interface with a set of noop methods and use this in newCoreMetrics if factory == nil.

thomasten:
> letting the callers of putState increment it has a great potential for people to forget

Agreed, this should be an atomic operation. And it is: the state is only supposed to be changed with advanceState. There's only one (recently introduced) exception in NewCore.

> the state of the Coordinator wouldn't be collected correctly at recovery

For metrics that equal some value of the program's state, it would be preferable to have a single source of truth (which should be the store) and let prometheus pull it on demand. Is this possible?

coordinator/core/core.go
katexochen: Since I've changed the `coordinatorState` metric to use `GaugeFunc` instead, I've only added the null object pattern for the `marbleAPI` metrics.

katexochen:
>  and let prometheus pull it on demand. Is this possible?

I've changed the `coordinatorState` metric to use `GaugeFunc`, which calls `getState` everytime it is collected, and added the null object pattern for the marbleAPI metrics.

thomasten (approved):
Sweet, thanks!
Sorry, didn't expect the null object pattern to be so many types and funcs. But I like.

m1ghtym0 (approved):
Thanks! Can you rebase and squash your commits?

PR #200, opened 2021-07-01
AB#1050 Rework stored data

daniel-weisse:
### Proposed changes
Rework the way data from the initial manifest is stored and handled throughout the coordinators lifecycle:
Rather than only storing newly generated data, once the initial manifest is set, all components of it are saved to the coordinators store. This includes packages, marbles, and TLS, as well as metadata for private and user-defined secrets.

This means that we no longer need to load the entire manifest and go through it if we want to access a specific resource, but can instead retrieve that resource directly from store.
Additionally this allows us to update values like a packages security version directly, without it affecting the fingerprint of the manifest.
This removes the need to store update manifests, and allows users to use update manifests containing only the packages they want to update, instead of also including all packages that have already been updated.

### Additional info
The original manifest is still needed for for 3 scenarios:
* Generating the manifests fingerprint.
* Reloading the components list of the core. This list is used to cache the names of all named resources stored in the core.
  This is done once after the Coordinator unsealed itself from a sealed state or after a manual recovery.
* Regenerating shared certificate secrets on a manifest update. Metadata containing information about the end of a certificates lifetime is changed when the secret is first created, creating the need for the raw cert specifications in this case.
We could store this information separately, but I don't think this would be the best solution, as in that case we would effectively store these secrets twice leading to a potentially confusing setup.


coordinator/core/core.go
```diff
@@ -79,6 +80,18 @@ const (
 	sKCoordinatorIntermediateKey  string = "coordinatorIntermediateKey"
 )
 
+// components is a list of the objects stored in the Coordinators store
```
m1ghtym0, line 83: The idea of components is essentially a cache that simplifies accessing all values of a certain type, right?
For a single instance Coordinator, this is fine. If we go down the route of replication this structure is essentially useless.
Not sure if it makes sense to introduce it now with replication as a feature on the horizon.

coordinator/core/clientapi.go
```diff
@@ -96,11 +97,39 @@ func (c *Core) SetManifest(ctx context.Context, rawManifest []byte) (map[string]
...
+		}
+	}
+	for k, v := range secrets {
+		if err := txdata.putSecret(k, v); err != nil {
+			return nil, err
+		}
+	}
+	// save metadata of private and user-defined secrets
+	for k, v := range manifest.Secrets {
+		if v.UserDefined || !v.Shared {
```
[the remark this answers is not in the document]
daniel-weisse: We dont actually save a secret here, but the metadata for creating the secrets. (Size of the key, x509 config for generating certificates etc.)
This information is loaded when we generated the secret for individual marbles and used to create the actual secret.

coordinator/core/core.go
daniel-weisse: Correct, otherwise we would need to load the manifest, go through it to learn what values exists, and then request them. This is essentially what the core does at the moment when loading users or secrets.

Regarding replication: As far as I understand it, at no point in time do we want to create new objects that are not documented in the manifest. This means that, once we have a manifest, we can generate a list of all components that will exists throughout the coordinators lifetime.
A newly duplicated coordinator instance has access to the same store, so, upon setup, the instance would request the manifest, and is then able to generate the same cache of component as any other instance of the coordinator.

thomasten: Moritz is right. And it's not only about replication, but also about code complexity and maintainability.

> otherwise we would need to load the manifest, go through it to learn what values exists

The idiomatic way is iterating over the keys of a KV store.

> at no point in time do we want to create new objects that are not documented in the manifest

I bet we will do this one day ;)

That said, the components thing is fine with me as an intermediate stage. But the long-term goal should be that the store is the single source of truth. Once we have this, new things will be easier to implement. We've already seen complexity reductions of existing features in the previous PRs and also in this PR.

m1ghtym0 (approved):

PR #201, opened 2021-07-01
Update Dockerfile

thomasten:
Installing linux-headers failed on my machine, but it's obsolete anyway, right?
I pushed the image to the registry. Can you pls test if it works?


3u13r (approved):
This PR is correct. 
Sadly, both our pushed image as well as the image I tested the changes with before I merged the sample PR fail with the same error. I am still investigating the root cause or what changed between my tests.

m1ghtym0:
@3u13r Should we keep this PR open until you found the bug, or should this be merged anyway?

3u13r:
IMO, since the current version could maybe lead to errors, we should merge this now. I will try to debug this some more today.

PR #203, opened 2021-07-05
AB#1061 Virtual manifest

daniel-weisse:
### Proposed changes
- Return the initial manifest over the coordinators REST API on GET requests to `/manifest`
  - Coordinator still send the signature and the general structure of the responds stays the same:
    ```go
    type manifestSignatureResp struct {
      ManifestSignature string
      Manifest          []byte
    }
    ```
  - This allows users who dont have access to the manifest at the start, to request the manifest at their own time directly from the coordinator.
- Change the `manifest get` command to mainly display the manifest instead of its signature
  - Using flags users can additionally display the signature, or create a virtual updated manifest by merging changes from the update log into the manifest.

### Additional info
- Not totally happy with the flag names. For now I am using `consolidate` / `c` to merge the updates into the manifest. Other options might be more intuitive.
- It could be possible to not return the manifest over the coordinators rest api, and instead move the virtual / consolidated manifest feature into another command. This would require users to supply the manifest themselves. The cli in this case only updates the manifest using an update log. This option does not seem the most user friendly in my opinion.

### DIFF
--- a/cli/cmd/manifestGet.go
+++ b/cli/cmd/manifestGet.go
@@ -50,19 +50,19 @@ Optionally get the manifests signature or merge updates into the displayed manif
 	}
 
 	cmd.Flags().BoolVarP(&signature, "signature", "s", false, "Set to additionally display the manifests signature")
-	cmd.Flags().BoolVarP(&consolidate, "consolidate", "c", false, "Set to merge updates into the displayed manifest")
+	cmd.Flags().BoolVarP(&consolidate, "display-update", "u", false, "Set to merge updates into the displayed manifest")
 	cmd.Flags().StringVarP(&output, "output", "o", "", "Save output to file instead of printing to stdout")
 	return cmd
 }
 
 // decodeManifest parses a base64 encoded manifest and optionally merges updates
-func decodeManifest(wantUpdate bool, encodedManifest, hostName string, cert []*pem.Block) (string, error) {
+func decodeManifest(consolidate bool, encodedManifest, hostName string, cert []*pem.Block) (string, error) {
 	manifest, err := base64.StdEncoding.DecodeString(encodedManifest)
 	if err != nil {
 		return "", err
 	}
 
-	if !wantUpdate {
+	if !consolidate {
 		return string(manifest), nil
 	}
 
--- a/cli/cmd/manifestGet.go
+++ b/cli/cmd/manifestGet.go
@@ -14,7 +14,7 @@ import (
 
 func newManifestGet() *cobra.Command {
 	var output string
-	var consolidate bool
+	var displayUpdate bool
 	var signature bool
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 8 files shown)


cli/cmd/manifestGet.go
```diff
@@ -1,39 +1,126 @@
...
-			fmt.Printf("Manifest signature: %s\n", string(response))
+			fmt.Println(manifest)
 			return nil
 		},
 		SilenceUsage: true,
 	}
-	cmd.Flags().StringVarP(&output, "output", "o", "", "Save singature to file instead of printing to stdout")
+
+	cmd.Flags().BoolVarP(&signature, "signature", "s", false, "Set to additionally display the manifests signature")
+	cmd.Flags().BoolVarP(&consolidate, "display-update", "u", false, "Set to merge updates into the displayed manifest")
```
m1ghtym0, line 53:
```suggestion
	cmd.Flags().BoolVarP(&displayUpdate, "display-update", "u", false, "Set to merge updates into the displayed manifest")
```

cli/cmd/manifestGet.go
```diff
@@ -1,39 +1,126 @@
...
 	"io/ioutil"
 
+	"github.com/edgelesssys/marblerun/coordinator/manifest"
 	"github.com/spf13/cobra"
+	"github.com/tidwall/gjson"
 )
 
 func newManifestGet() *cobra.Command {
 	var output string
+	var consolidate bool
```
m1ghtym0, line 17:
```suggestion
	var displayUpdate bool
```

cli/cmd/manifestGet.go
```diff
@@ -1,39 +1,126 @@
...
+
+	updated, err := json.Marshal(baseManifest)
+	if err != nil {
+		return "", err
+	}
+	var removeTarget map[string]interface{}
+	if err := json.Unmarshal(updated, &removeTarget); err != nil {
+		return "", err
+	}
+	removeNil(removeTarget)
```
m1ghtym0, line 98: Why are we doing this?

m1ghtym0 (changes requested):

cli/cmd/manifestGet.go
daniel-weisse: Prevents all the unset values from being displayed:

Output from `manifest get $MARBLERUN -u` with nil removal:
```javascript
{
    "Marbles": {
        "hello": {
            "MaxActivations": 0,
            "Package": "world",
            "Parameters": {
                "Argv": ["foo", "bar"]
            }
        }
    },
    "Packages": {
        "world": {
            "Debug": true,
            "SignerID": "",
            "UniqueID": "e7a6a1f81bfce87067d8f6e2aada15a1053b233fd87fd533b13492617a40646a"
        }
    }
}
```

and without:
```javascript
{
    "Packages": {
        "world": {
            "Debug": true,
            "UniqueID": "e7a6a1f81bfce87067d8f6e2aada15a1053b233fd87fd533b13492617a40646a",
            "SignerID": "",
            "ProductID": null,
            "SecurityVersion": null
        }
    },
    "Infrastructures": null,
    "Marbles": {
        "hello": {
            "Package": "world",
            "MaxActivations": 0,
            "Parameters": {
                "Argv": ["foo", "bar"]
            },
            "TLS": null
        }
    },
    "Users": null,
    "Clients": null,
    "Secrets": null,
    "RecoveryKeys": null,
    "Roles": null,
    "TLS": null
}
```

m1ghtym0 (approved):

m1ghtym0:
Please squash, than we can merge this

PR #206, opened 2021-07-07
AB#1064 Replace components with store iterator

daniel-weisse:
### Proposed changes
Replace the components structure introduced in https://github.com/edgelesssys/marblerun/pull/200 with an Iterator function for the store.
This means that all data, and information about said data, is kept in the store, providing full separation of core and store.

### Additional info
The Iterator takes a string (a prefix for a given type) and returns a list of all keys in the store for this prefix.
So if we wanted to get an iterator to request users, we call the iterator function with the `requestUser` string, resulting in a list containing the keys the values for all users are saved under:
```
userList: ["user:alice", "user:bob", "user:oscar"]
```

The storewrapper takes care of removing any prefixes from the resource names, if they are present. 
So it does not matter if the implementation of a store returns the prefix or chooses to omit it, meaning users of the storewrapper will always receive the list without prefix, no matter what.


coordinator/core/marbleapi.go
```diff
@@ -92,7 +92,7 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
 	}
 
 	// Generate unique (= per marble) secrets
-	privateSecrets, err := c.data.getSecretMap(c.cmp.privateSecrets)
+	privateSecrets, err := c.data.getSecretMap(true)
```
m1ghtym0, line 95: Can we simplify this by not distinguishing secrets in the store?
Instead, we could handle different secrets in `generateSecrets` and only generate the private ones.
This would simplify the code quite a bit.

coordinator/store/store.go
```diff
@@ -16,6 +16,8 @@ type Store interface {
 	Get(string) ([]byte, error)
 	// Put saves a value to store by key
 	Put(string, []byte) error
+	// Iterator returns a list of keys with a given prefix
+	Iterator(string) ([]string, error)
```
m1ghtym0, line 20: Not sure what's the idiomatic golang way here.
Iterator could be an interface that has `next` and `hasNext` functions.
Using a list we need to assume we can hold all data of the store in memory.
Should be ok in most cases.
Got an opinion @thomasten?

coordinator/core/marbleapi.go
daniel-weisse: Since the `Shared` flag is stored with the secret it should be possible to generate the list at runtime by requesting the secrets and then checking whether or not they are private.
And since for a marble activation we need to request all shared secrets anyways there should be no performance difference.

daniel-weisse:
While working on changing the way private secrets are handled i noticed a rather significant bug:

The way certificate secrets are stored causes any information in the `Cert` field of the manifest to be dropped once it is marshaled to JSON.
This is not a problem for shared secrets (the information in `Cert` is used to create a valid x509 certificate containing that information).
Private secrets however lose all metadata, resulting in all private certificates to only contain default information. 

I'm looking for a way keep this information in an efficient way, this might also help for regenerating shared certificates.

coordinator/core/core.go
```diff
@@ -554,7 +538,14 @@ func (c *Core) generateCertificateForSecret(secret manifest.Secret, parentCertif
...
 		template.NotAfter = time.Now().AddDate(0, 0, int(secret.ValidFor))
 	} else if secret.ValidFor != 0 {
-		return manifest.Secret{}, errors.New("ambigious certificate validity duration, both NotAfter and ValidFor are specified")
+		if uint(template.NotAfter.Sub(oldBefore).Hours()/24) != secret.ValidFor {
+			// The duration of NotBefore to NotAfter did not match ValidFor, meaning both were specified in the manifest
+			return manifest.Secret{}, errors.New("ambigious certificate validity duration, both NotAfter and ValidFor are specified")
+		}
+		// reset expiration date for private secrets
+		if !secret.Shared {
+			template.NotAfter = time.Now().AddDate(0, 0, int(secret.ValidFor))
```
m1ghtym0, line 547: That way we're handling NotBefore and NotAfter implicitly as ValidFor.
However, if I set `NotAfter` in the manifest I would not expect, this to be a dynamic value.
I'd rather explicitly mark `ValidFor` as an overwrite for `NotBefore`/`NotAfter` or always throw an error.

coordinator/store/store.go
m1ghtym0: Please use a `next`, `hasNext` interface.

m1ghtym0 (changes requested):

coordinator/core/core.go
daniel-weisse: If `ValidFor` is set in the manifest, the expiration date will always be that specific date. I think having an option to do this is important.
The case we see here, where `NotAfter` gets dynamically created using `ValidFor`, wont happen if only `ValidFor` is set.

m1ghtym0: The scenario I'm thinking here is the following:
* I have a manifest with a `ValidFor` value set (e.g. 30 days)
* I want to change my deployment to be only valid in a certain time frame. (e.g. 01. August - 31. August) and set `NotBefore` and `NotAfter` accordingly. However, I forget to delete the `ValidFor` set.
* For some reason a pod restarts on 10. August
* I'd expect this pod to have a certificate valid until 31. August, but it's valid from 10. August - 09. September

This is what I meant with implicit

coordinator/store/stdstore.go
```diff
@@ -144,6 +156,18 @@ func (t *transaction) Put(request string, requestData []byte) error {
...
+// Iterator returns an iterator for all keys in the transaction with a given prefix
+func (t *transaction) Iterator(prefix string) (Iterator, error) {
+	keys := make([]string, 0)
+	for k := range t.data {
+		if strings.HasPrefix(k, prefix) {
+			keys = append(keys, k)
+		}
+	}
+
+	return &StdIterator{-1, keys, nil}, nil
```
m1ghtym0, line 168:
```suggestion
	return &StdIterator{0, keys, nil}, nil
```

coordinator/store/stdstore.go
```diff
@@ -159,3 +183,33 @@ func (t *transaction) Rollback() {
...
+	}
+	return i.keys[i.idx]
+}
+
+// HasNext implements the Iterator interface
+func (i *StdIterator) HasNext() bool {
+	if (i.idx + 1) < len(i.keys) {
+		return true
+	}
+	return false
```
m1ghtym0, line 209:
```suggestion
	if i.idx < len(i.keys) {
		return true
	}
	return false
```

coordinator/store/stdstore.go
```diff
@@ -159,3 +183,33 @@ func (t *transaction) Rollback() {
...
+}
+
+// Next implements the Iterator interface
+func (i *StdIterator) GetNext() string {
+	i.idx++
+	if i.idx >= len(i.keys) {
+		i.err = fmt.Errorf("index out of range [%d] with length %d", i.idx, len(i.keys))
+		return ""
+	}
+	return i.keys[i.idx]
```
m1ghtym0, line 201:
```suggestion
	if i.idx >= len(i.keys) {
		return "",  fmt.Errorf("index out of range [%d] with length %d", i.idx, len(i.keys))
	}
	val := i.keys[i.idx]
	i.idx++
	return val, nil
```

coordinator/store/stdstore.go
```diff
@@ -159,3 +183,33 @@ func (t *transaction) Rollback() {
...
+
+// StdIterator is the standard Iterator implementation
+type StdIterator struct {
+	idx  int
+	keys []string
+	err  error
+}
+
+// Next implements the Iterator interface
+func (i *StdIterator) GetNext() string {
```
m1ghtym0, line 195:
```suggestion
func (i *StdIterator) GetNext() (string, error) {
```

coordinator/store/stdstore.go
```diff
@@ -159,3 +183,33 @@ func (t *transaction) Rollback() {
...
+	if (i.idx + 1) < len(i.keys) {
+		return true
+	}
+	return false
+}
+
+// Error implements the Iterator interface
+func (i *StdIterator) Error() error {
+	return i.err
+}
```
m1ghtym0, line 215:
```suggestion
```

coordinator/core/clientapi.go
```diff
@@ -241,18 +251,25 @@ func (c *Core) GetUpdateLog(ctx context.Context) (string, error) {
...
 	// Check if a supplied client cert matches the supplied ones from the manifest stored in the core
 	// NOTE: We do not use the "correct" X.509 verify here since we do not really care about expiration and chain verification here.
 	for _, suppliedCert := range clientCerts {
-		for _, userName := range c.cmp.users {
-			user, err := c.data.getUser(userName)
+		for userIter.HasNext() {
+			user, err := c.data.getUser(userIter.GetNext())
 			if err != nil {
 				return nil, err
 			}
```
m1ghtym0, line 265:
```suggestion
			userName, err := userIter.GetNext()
			if err != nil {
				return nil, err
			}
			user, err := c.data.getUser(userName)
			if err != nil {
				return nil, err
			}
```

coordinator/core/clientapi.go
```diff
@@ -241,18 +251,25 @@ func (c *Core) GetUpdateLog(ctx context.Context) (string, error) {
...
 			if suppliedCert.Equal(user.Certificate()) {
 				return user, nil
 			}
 		}
+		if userIter.Error() != nil {
+			return nil, err
+		}
```
m1ghtym0, line 272:
```suggestion
```

coordinator/core/marbleapi.go
```diff
@@ -173,15 +170,20 @@ func (c *Core) verifyManifestRequirement(tlsCert *x509.Certificate, certQuote []
...
+		if !infraIter.HasNext() {
 			if err := c.qv.Validate(certQuote, tlsCert.Raw, pkg, quote.InfrastructureProperties{}); err != nil {
 				return status.Errorf(codes.Unauthenticated, "invalid quote: %v", err)
 			}
 		} else {
 			infraMatch := false
-			for _, infraName := range c.cmp.infrastructures {
-				infra, err := c.data.getInfrastructure(infraName)
+			for infraIter.HasNext() {
+				infra, err := c.data.getInfrastructure(infraIter.GetNext())
```
m1ghtym0, line 186:
```suggestion
			for infraIter.HasNext() {
				infraName, err := infraIter.GetNext()
				if err != nil {
					return err
				}
				infra, err := c.data.getInfrastructure(infraName)
```

coordinator/core/storewrapper.go
```diff
@@ -39,9 +39,34 @@ type storeWrapper struct {
...
+}
+
+func (i iteratorWrapper) HasNext() bool {
+	return i.iterator.HasNext()
+}
+
+func (i iteratorWrapper) Error() error {
+	return i.iterator.Error()
+}
+
```
m1ghtym0, line 63:
```suggestion
```

coordinator/store/stdstore.go
```diff
@@ -60,6 +59,19 @@ func (s *StdStore) Put(request string, requestData []byte) error {
...
+// For an empty prefix this is an iterator for all keys in StdStore
+func (s *StdStore) Iterator(prefix string) (Iterator, error) {
+	keys := make([]string, 0)
+	for k := range s.data {
+		if strings.HasPrefix(k, prefix) {
+			keys = append(keys, k)
+		}
+	}
+
+	return &StdIterator{-1, keys, nil}, nil
```
m1ghtym0, line 72:
```suggestion
	return &StdIterator{0, keys, nil}, nil
```

coordinator/store/stdstore.go
```diff
@@ -159,3 +183,33 @@ func (t *transaction) Rollback() {
 		t.store.txmux.Unlock()
 	}
 }
+
+// StdIterator is the standard Iterator implementation
+type StdIterator struct {
+	idx  int
+	keys []string
+	err  error
```
m1ghtym0, line 191:
```suggestion
```

coordinator/store/store.go
```diff
@@ -24,12 +26,24 @@ type Transaction interface {
...
 
+// Iterator is an iterator for the store
+type Iterator interface {
+	// Returns the next element of the iterator
+	GetNext() string
+	// HasNext returns true if there is at least one more item after the current position
+	HasNext() bool
+	// Error returns errors encountered during iteration
+	Error() error
+}
```
m1ghtym0, line 45:
```suggestion
type Iterator interface {
	// Returns the next element of the iterator
	GetNext() (string, error)
	// HasNext returns true if there is at least one more item after the current position
	HasNext() bool
}
```

coordinator/store/stdstore_test.go
```diff
@@ -46,24 +41,76 @@ func TestStdStore(t *testing.T) {
...
+		"value:1":   {0x00, 0x11},
+		"something": {0x00},
+	}
+
+	iter, err := store.Iterator("test")
+	assert.NoError(err)
+	idx := 0
+	for iter.HasNext() {
+		idx++
+		assert.Contains(iter.GetNext(), "test:")
```
m1ghtym0, line 68:
```suggestion
		val, err := iter.GetNext()
		assert.NoError(err)
		assert.Contains(val, "test:")
```

coordinator/store/stdstore_test.go
```diff
@@ -46,24 +41,76 @@ func TestStdStore(t *testing.T) {
...
+	}
+	assert.EqualValues(3, idx)
+	assert.NoError(iter.Error())
```
m1ghtym0, line 71:
```suggestion
```

coordinator/store/stdstore_test.go
```diff
@@ -46,24 +41,76 @@ func TestStdStore(t *testing.T) {
...
+		assert.Contains(iter.GetNext(), "value:")
+	}
+	assert.EqualValues(1, idx)
+	assert.NoError(iter.Error())
+
+	iter, err = store.Iterator("")
+	assert.NoError(err)
+	idx = 0
+	for iter.HasNext() {
+		iter.GetNext()
```
m1ghtym0, line 87:
```suggestion
		_, err = iter.GetNext()
		assert.NoError(err)
```

coordinator/core/storewrapper.go
```diff
@@ -181,20 +206,26 @@ func (s storeWrapper) putSecret(secretName string, secret manifest.Secret) error
...
-	for _, name := range secretNames {
+	secretMap := map[string]manifest.Secret{}
+	for iter.HasNext() {
 		// all secrets (user-defined and private only as uninitialized placeholders) are set with the initial manifest
 		// if we encounter an error here something went wrong with the store, or the provided list was faulty
+		name := iter.GetNext()
 		secretMap[name], err = s.getSecret(name)
 		if err != nil {
 			return nil, err
 		}
```
m1ghtym0, line 224:
```suggestion
		name, err := iter.GetNext()
		if err != nil {
			return nil, err
		}
		secretMap[name], err = s.getSecret(name)
		if err != nil {
			return nil, err
		}
```

m1ghtym0 (changes requested):

m1ghtym0 (approved):

Issue #216, opened 2021-07-20
[Build] Parallel build fails


dimakuv:
### Issue description

When trying to build Marblerun from sources, I follow the [BUILD guide](https://github.com/edgelesssys/marblerun/blob/master/BUILD.md) but replace `make` with `make -j` (enables parallel build on my multi-core machine).

This breaks the build like this:
```
..Scanning dependencies of target sign-coordinator
+++[ 60%] Generating private.pem, public.pem
+
.+Generating RSA private key, 3072 bit long modulus (2 primes)
+.++
e is 3 (0x03)
.unable to load Private Key
140642674569664:error:0909006C:PEM routines:get_name:no start line:../crypto/pem/pem_lib.c:745:Expecting: ANY PRIVATE KEY
.CMakeFiles/sign-marble-test.dir/build.make:90: recipe for target 'private.pem' failed
make[2]: *** [private.pem] Error 1
make[2]: *** Deleting file 'private.pem'
CMakeFiles/Makefile2:139: recipe for target 'CMakeFiles/sign-marble-test.dir/all' failed
make[1]: *** [CMakeFiles/sign-marble-test.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
.....++++
...................................++++
e is 3 (0x03)
Can't open private.pem for reading, No such file or directory
140645527962048:error:02001002:system library:fopen:No such file or directory:../crypto/bio/bss_file.c:72:fopen('private.pem',
'r')
140645527962048:error:2006D080:BIO routines:BIO_new_file:no such file:../crypto/bio/bss_file.c:79:
unable to load Private Key
CMakeFiles/sign-coordinator.dir/build.make:90: recipe for target 'private.pem' failed
```

It looks like the dependency on auto-generated `private.pem` is not correctly written in one of the Makefiles (or in CMake files).

### To reproduce

Follow the [BUILD guide](https://github.com/edgelesssys/marblerun/blob/master/BUILD.md) but replace `make` with `make -j`.

### Expected behavior

The parallel build should succeed.

### Environment:
 - Marblerun version: latest master, commit 3780c0eae950fdac9255018a8e328f9f90f7ea2b
 - Edgeless RT version: v0.2.5
 - Go version: go version go1.16.3 linux/amd64
 - Minikube version: N/A

PR #218, opened 2021-07-20
Fix parallel build

m1ghtym0:
### Proposed changes
- Fix dependencies on `private.pem`
- My guess is that OpenSSL creates the file before generating the key, triggering the dependent Makefile targets.

### Related issue
- https://github.com/edgelesssys/marblerun/issues/216

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,10 @@ add_custom_command(
   COMMAND openssl genrsa -out private.pem -3 3072
   COMMAND openssl rsa -in private.pem -pubout -out public.pem)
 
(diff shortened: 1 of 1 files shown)


dimakuv:
For what it's worth, I tested this PR quickly on my machine (with `make -j` on 12 cores). Previously it failed, now it worked all 5 times. Legit.

daniel-weisse (approved):

thomasten (approved):
Problem seems to be that the custom command is invoked twice and concurrently if the parallel built targets depend on the same non-existing file. Depending on a target prevents this behavior.

PR #220, opened 2021-07-20
Remove sealed_key if no sealed state was found

daniel-weisse:
### Proposed changes
- Remove sealed key (if one exists) during `AESGCMSealer.Unseal()` if no sealed state is found


m1ghtym0 (approved):

thomasten (changes requested):
block

thomasten (approved):

PR #233, opened 2021-08-11
Change Marblerun to MarbleRun

danibentrup:
### Proposed changes
-
-


samples/helloworld/enclave.json
```diff
@@ -0,0 +1,8 @@
+{
+ "exe": "hello",
```
Nirusu, line 2: Is this file left from testing?

samples/helloc++/README.md
```diff
@@ -1,7 +1,7 @@
...
 
-**Note:** You can run this example on any hardware by simulating the enclave through setting `OE_SIMULATION=1` as environment variable. This might help you to get started with with the development of confidential apps. However, please notice that this bypasses any security. Detailed information on how to develope secure Marbles can be found in [Marbleruns documentation](https://docs.edgeless.systems/marblerun/#/workflows/add-service).
+**Note:** You can run this example on any hardware by simulating the enclave through setting `OE_SIMULATION=1` as environment variable. This might help you to get started with with the development of confidential apps. However, please notice that this bypasses any security. Detailed information on how to develope secure Marbles can be found in [MarbleRuns documentation](https://docs.edgeless.systems/marblerun/#/workflows/add-service).
```
Nirusu, line 4:
```suggestion
**Note:** You can run this example on any hardware by simulating the enclave through setting `OE_SIMULATION=1` as environment variable. This might help you to get started with with the development of confidential apps. However, please notice that this bypasses any security. Detailed information on how to develop secure Marbles can be found in [MarbleRun's documentation](https://docs.edgeless.systems/marblerun/#/workflows/add-service).
```

cli/cmd/check.go
```diff
@@ -17,8 +17,8 @@ func newCheckCmd() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "check",
-		Short: "Check the status of Marbleruns control plane",
-		Long:  `Check the status of Marbleruns control plane`,
+		Short: "Check the status of MarbleRuns control plane",
+		Long:  `Check the status of MarbleRuns control plane`,
```
Nirusu, line 21:
```suggestion
		Long:  `Check the status of MarbleRun's control plane`,
```

cli/cmd/check.go
```diff
@@ -17,8 +17,8 @@ func newCheckCmd() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "check",
-		Short: "Check the status of Marbleruns control plane",
-		Long:  `Check the status of Marbleruns control plane`,
+		Short: "Check the status of MarbleRuns control plane",
```
Nirusu, line 20:
```suggestion
		Short: "Check the status of MarbleRun's control plane",
```

samples/helloworld/README.md
```diff
@@ -1,7 +1,7 @@
...
 
-**Note:** You can run this example on any hardware by simulating the enclave through setting `OE_SIMULATION=1` as environment variable. This might help you to get started with with the development of confidential apps. However, please notice that this bypasses any security. Detailed information on how to develope secure Marbles can be found in [Marbleruns documentation](https://docs.edgeless.systems/marblerun/#/workflows/add-service).
+**Note:** You can run this example on any hardware by simulating the enclave through setting `OE_SIMULATION=1` as environment variable. This might help you to get started with with the development of confidential apps. However, please notice that this bypasses any security. Detailed information on how to develope secure Marbles can be found in [MarbleRuns documentation](https://docs.edgeless.systems/marblerun/#/workflows/add-service).
```
Nirusu, line 4:
```suggestion
**Note:** You can run this example on any hardware by simulating the enclave through setting `OE_SIMULATION=1` as environment variable. This might help you to get started with with the development of confidential apps. However, please notice that this bypasses any security. Detailed information on how to develop secure Marbles can be found in [MarbleRun's documentation](https://docs.edgeless.systems/marblerun/#/workflows/add-service).
```

Nirusu (approved):
LGTM for the most part. Just not really happy that this is actually a breaking change for the Secrets feature... Not sure if building another wrapper for "Marblerun" here might actually be not that bad. I know we usually do not keep legacy functionality yet, however I think that having to spell "MarbleRun" explicitly with this capitalization is a particularity that might be weird to deal with in general. 

But yeah, despite this some minor things I notices when I skimmed through. And other than that, the change is fine :)

samples/helloworld/enclave.json
m1ghtym0: Yes, accidental commit.

m1ghtym0:
@Nirusu Is your suggestion a wrapper that deals with all "spellings" of MarbleRun in the manifest secrets?

Nirusu:
Not every type, just "Marblerun" or "marblerun" maybe, given the first one for backward compatibility, and the second one since we also use "snake_case" in several of our examples for the type behind "Marblerun", and this would make things more consistent in terms of style?

Though, this can also be omitted if it makes even less sense to just support certain spelling. I think manually creating a type called "maRBleRun" would be a very strange thing to do :D

Nirusu:
Actually, I think "marblerun" is not even allowed by Go's standards as a type name given that lower-case start implies "private" variables. Same with other casings.

So mh, maybe just a wrapper type for the old name (avoids many redundant changes, though this means "legacy code" even if it is just a wrapper), case-insensitive template parsing (if that would be possible, I don't think it is) or just go ahead and break things.

I think the first one is acceptable, but you can decide.

m1ghtym0:
I see your point.
I don't think we're at the point yet where we should think about keeping this for backward compatibility.
We're using camel case for most stuff in the manifest which is consistent with the Go convention. We should actually get rid of the snake case.
Since all lowercase is not an option, using "Marblerun" would be weird.
After all, from now on there is only MarbleRun no more Marblerun.
Consistency is key here IMO. If we start using one or the other in different scenarios we will eventually get people confused.

Nirusu:
>  We should actually get rid of the snake case.
I agree with this.

> If we start using one or the other in different scenarios we will eventually get people confused.
Personally I don't think that two types mapping the same thing is particularly bad or very confusing. But yeah, having particular wrappers is also dumb. 

I am fine with the change. It's just going to be a pain to change this everywhere we defined manifests :)

m1ghtym0:
We need to adjust the manifest in all external examples and the documentation when this is released.
I'll add a ticket for that.

coordinator/core/clientapi_test.go
```diff
@@ -428,6 +429,11 @@ func TestGetSecret(t *testing.T) {
 	admin, err := c.data.getUser("admin")
 	assert.NoError(err)
 
+	fmt.Println("Cert")
+	fmt.Printf(base64.RawStdEncoding.EncodeToString(secret2.Cert.Raw) + "\n")
+	fmt.Println("Key")
+	fmt.Printf(base64.RawStdEncoding.EncodeToString(secret2.Private) + "\n")
```
daniel-weisse, line 435:
```suggestion
```
Can this be removed?

daniel-weisse (approved):

PR #235, opened 2021-08-16
AB#1166 Fix parsing of raw secrets

daniel-weisse:
### Proposed changes
- Fixes a problem where secrets injected as files or env variables using the `raw` encoding resulted in an error when sending the data from the Coordinator to a Marble due to the data having an extremely high likelihood of being non UTF-8.
- Define `Files` and `Env` as `bytes` in the rpc code to fix this issue.
  - This is a breaking change for all Marbles compiled before this change is merged into master and released as a new version!
  - EGo will need a new release based on this new MarbleRun version
- This also decouples the `Files` and `Env` entries in the manifest from the same in the rpc code, saving us a few nil pointer checks, and allows easier changes to these entries in the future.

### Additional info
- Due to the issue being with our grpc implementation this problem is not caught with simple unit tests. However, our integration test works for this and now includes `raw` secrets in the manifest.
- I have also added the recommended embedding of `Unimplemented*Server` to allow for forward compatible implementations, if we ever decide to add additional functionality to the Marble API


coordinator/core/marbleapi_test.go
```diff
@@ -163,24 +163,28 @@ func (ms *marbleSpawner) newMarble(marbleType string, infraName string, shouldSu
 	params := resp.GetParameters()
 	// Validate Files
 	if marble.Parameters.Files != nil {
-		ms.assert.Equal(marble.Parameters.Files, params.Files)
+		for k, v := range marble.Parameters.Files {
+			ms.assert.EqualValues(v, resp.Parameters.Files[k])
+		}
+		//ms.assert.Equal(marble.Parameters.Files, params.Files)
+		//ms.assert.EqualValues(marble.Parameters.Files, params.Files)
```
m1ghtym0, line 170: Can these be removed?

coordinator/core/marbleapi_test.go
```diff
@@ -545,14 +549,12 @@ func (ms *marbleSpawner) shortMarbleActivation(marbleType string, infraName stri
...
 	// Validate Files
-	if marble.Parameters != nil {
-		if marble.Parameters.Files != nil {
-			ms.assert.Equal(marble.Parameters.Files, params.Files)
-		}
-		// Validate Argv
-		if marble.Parameters.Argv != nil {
-			ms.assert.Equal(marble.Parameters.Argv, params.Argv)
-		}
+	if marble.Parameters.Files != nil {
```
m1ghtym0, line 552: Is the check for `nil` obsolete?

m1ghtym0 (approved):

coordinator/core/marbleapi_test.go
daniel-weisse: Good catch, the nil check is obsolete, but we need to compare the file entries individually instead of comparing maps

m1ghtym0: Ah, so gRPC is making sure a request can not have an empty `Parameters` field?

daniel-weisse: No, this is a problem with the testing framework.
We compare `map[string]string` to `map[string][]bytes`, which always returns false.
We fix this by comparing the map values one by one instead all at the same time, since comparing `string` and `[]byte` evaluates as true, if the values are the same

m1ghtym0 (approved):

PR #236, opened 2021-08-17
AB#1138 / AB#1139: Add escaped PEM as secret encoding

Nirusu:
### Proposed changes
- Add a new type `pemEscaped` for secrets to inject PEM secrets into e.g. JSON configuration so we can replace such a secret definition:
```json
"ca": "-----BEGIN CERTIFICATE-----\n{{ base64 .Marblerun.RootCA.Cert }}\n-----END CERTIFICATE-----\n"
```

With this one:
```json
"ca": "{{ pemEscaped .Marblerun.RootCA.Cert }}"
```

Normal `pem` would fail here, as JSON expects new lines to be escaped and does not have a "raw string" type, unfortunately.

### Additional info
- Personally, I do not quite like the name `pemEscaped`, however I also did not want to call it `pemJSON` as it might also be useful in other types. Let me know if you have better suggestions.

### DIFF
--- a/coordinator/core/marbleapi_test.go
+++ b/coordinator/core/marbleapi_test.go
@@ -374,6 +374,10 @@ func TestParseSecrets(t *testing.T) {
 		Secrets:   testSecrets,
 	}
 
+	type testEscapedPemForJson struct {
+		Certificate string `json:"certificate"`
+	}
+
 	// Test all formats, pem should fail for raw/symmetric secrets
 	parsedSecret, err := parseSecrets("{{ raw .Secrets.mysecret }}", testWrappedSecrets)
 	require.NoError(err)
@@ -384,6 +388,9 @@ func TestParseSecrets(t *testing.T) {
 	assert.EqualValues("000102030405060708090a0b0c0d0e0f", parsedSecret)
 
 	_, err = parseSecrets("{{ pem .Secrets.mysecret }}", testWrappedSecrets)
+	assert.Error(err)
+
+	_, err = parseSecrets("{{ pemEscaped .Secrets.anothercoolsecret }}", testWrappedSecrets)
 	assert.Error(err)
 
 	parsedSecret, err = parseSecrets("{{ base64 .Secrets.mysecret }}", testWrappedSecrets)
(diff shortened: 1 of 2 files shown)


daniel-weisse (approved):

m1ghtym0 (approved):
`pemEscaped` is ok IMO. At least I don't have a better name right now;-)
Please add it to the docs as well.

thomasten:
Do we need this feature or could we also do the escaping of `pem .Marblerun.RootCA.Cert` within the Go template itself? I'm not so familiar with templates.
In general I'd prefer a more general approach instead of introducing a new type. (But only if the UX is okay.)

m1ghtym0:
The problem is that we don't want to escape `pem` in general.
But I agree with Thomas that a more general solution that would allow for escaping strings in the templates would be better.
Not sure if that is doable with the Golang templates though?

```
"ca": "{{ escaped  pem .Marblerun.RootCA.Cert }}"
"ca": "{{ escaped  {{ pem .Marblerun.RootCA.Cert }} }}"
```

Nirusu:
Would have to look this up and experiment, the syntax is cleaner... But I don't get on how we should escape the other types. They usually don't have any new lines anyway?

thomasten:
You can, e.g., do `{{ js pem .Marblerun.RootCA.Cert }}`. js is predefined and "returns the escaped JavaScript equivalent". So I'd suggest adding `escapeNewline`. Or maybe add nothing because `js` already does the job we need?

Nirusu:
Does prepending `js` actually work?

This would be ideal, but when I try this, I am greated with this lovely error:
```
template: data:10:17: executing "data" at <pem>: wrong number of args for pem: want 1 got 0
```

With the following definition:
```json
"ca": "{{ js pem .Marblerun.RootCA.Cert }}"
```

thomasten:
pls try
```
"ca": "{{ js (pem .Marblerun.RootCA.Cert) }}"
```

Nirusu:
This seems to work, great! Okay, in this case I guess the extra encoding is redundant and not needed.

PR #237, opened 2021-08-17
AB#1164 Enhanced files and env declarations

daniel-weisse:
### Proposed changes
Allow users to specify the encoding type of files and env in the manifest. See https://github.com/edgelesssys/marblerun/pull/231 for reference
Users may still define files or environment variables as a single string.

If not defined as a string these options are available:
  * `Data`: The content of the file
  * `Encoding`: The encoding of `Data` as it is written in the manifest. Currently supported are `base64`, `hex` and `string`.
    * `base64` marked content is decoded from base64 before it is passed to a Marble to be saved as a file/env var
    * `hex` is the same as `base64`, but content is decoded from hex
    * `string` causes no change to the input data
    * Spelling of `Encoding` is case insensitive. `string`, `String` and `sTrInG` are all valid spellings
  * `NoTemplates`: If this flag is set content in `Data` is not processed by the template engine and passed to a Marble as is. This means no secrets or other values are injected into the file at runtime.
 

In a manifest the file declaration can look similar to the following:
```Javascript
"Files": {
  "/file/from_string.txt": "Foo",
  "/file/from_struct.txt": {
    "Encoding": "string",
    "Data": "SGVsbG8gV29ybGQ=",
    "NoTemplates": true
  },
  "/file/from_base64.txt": {
    "Encoding": "base64",
    "Data": "SGVsbG8gV29ybGQ=",
    "NoTemplates": true
  }
}
```
Here `/file/from_struct.txt` and `/file/from_base64.txt` both have the same value set for `Data`, but because they declared different encoding types `/file/from_struct.txt` will contain `SGVsbG8gV29ybGQ=` when it is passed to a Marble, while  the content of `/file/from_base64.txt` will first get decoded, resulting in a file containing `Hello World`.

### DIFF
--- a/coordinator/core/marbleapi.go
+++ b/coordinator/core/marbleapi.go
@@ -17,6 +17,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"math"
+	"strings"
 	"text/template"
 	"time"
 
@@ -279,20 +280,34 @@ func customizeParameters(params manifest.Parameters, specialSecrets reservedSecr
 		Secrets:   userSecrets,
 	}
 
+	var err error
+	var newValue string
+
 	// replace placeholders in files
 	for path, data := range params.Files {
-		newValue, err := parseSecrets(data, secretsWrapped)
-		if err != nil {
-			return nil, err
+		if data.NoTemplates {
+			newValue = data.Data
+		} else {
+			newValue, err = parseSecrets(data.Data, secretsWrapped)
+			if err != nil {
+				return nil, err
+			}
 		}
 
 		customParams.Files[path] = []byte(newValue)
 	}
 
 	for name, data := range params.Env {
-		newValue, err := parseSecrets(data, secretsWrapped)
-		if err != nil {
-			return nil, err
+		if data.NoTemplates {
+			newValue = data.Data
+		} else {
+			if strings.Contains(data.Data, string([]byte{0x00})) {
+				return nil, fmt.Errorf("environment variable: %s: content contains null bytes", name)
+			}
+			newValue, err = parseSecrets(data.Data, secretsWrapped)
+			if err != nil {
+				return nil, err
+			}
 		}
 
 		customParams.Env[name] = []byte(newValue)
@@ -456,9 +471,9 @@ func (c *Core) setTTLSConfig(marble manifest.Marble, specialSecrets reservedSecr
 		return err
 	}
 	if marble.Parameters.Env == nil {
-		marble.Parameters.Env = make(map[string]string)
+		marble.Parameters.Env = make(map[string]manifest.File)
 	}
-	marble.Parameters.Env["MARBLE_TTLS_CONFIG"] = string(ttlsConfJSON)
+	marble.Parameters.Env["MARBLE_TTLS_CONFIG"] = manifest.File{Data: string(ttlsConfJSON), Encoding: "UTF-8"}
 
 	return nil
 }
--- a/coordinator/core/marbleapi.go
+++ b/coordinator/core/marbleapi.go
@@ -473,7 +473,7 @@ func (c *Core) setTTLSConfig(marble manifest.Marble, specialSecrets reservedSecr
 	if marble.Parameters.Env == nil {
 		marble.Parameters.Env = make(map[string]manifest.File)
 	}
-	marble.Parameters.Env["MARBLE_TTLS_CONFIG"] = manifest.File{Data: string(ttlsConfJSON), Encoding: "UTF-8"}
+	marble.Parameters.Env["MARBLE_TTLS_CONFIG"] = manifest.File{Data: string(ttlsConfJSON), Encoding: "string"}
 
 	return nil
 }
--- a/coordinator/manifest/manifest_test.go
+++ b/coordinator/manifest/manifest_test.go
@@ -24,7 +24,7 @@ func TestFile(t *testing.T) {
 {
 	"string": "helloworld",
 	"string_struct": {
-		"Encoding": "utf-8",
+		"Encoding": "string",
 		"NoTemplates": false,
 		"Data": "foo"
 	},
@@ -34,11 +34,15 @@ func TestFile(t *testing.T) {
 		"Data": "YmFy"
 	},
 	"base64_value": {
-		"Encoding": "utf-8",
+		"Encoding": "string",
 		"Data": "YmFy"
 	},
+	"hex": {
+		"Encoding": "hex",
+		"Data": "4d6172626c6552756e"
+	},
 	"no_templates": {
-		"Encoding": "utf-8",
+		"Encoding": "string",
 		"NoTemplates": true,
 		"Data": "{{ string .Secrets.symmetric_key_shared }}"
 	}
@@ -50,10 +54,11 @@ func TestFile(t *testing.T) {
 	err := json.Unmarshal(dataJSON, &testFiles)
 	require.NoError(err)
 	assert.Equal("helloworld", testFiles["string"].Data)
-	assert.Equal("UTF-8", testFiles["string"].Encoding)
+	assert.Equal("string", testFiles["string"].Encoding)
 	assert.Equal("foo", testFiles["string_struct"].Data)
 	assert.Equal("bar", testFiles["base64"].Data)
 	assert.Equal("YmFy", testFiles["base64_value"].Data)
+	assert.Equal("MarbleRun", testFiles["hex"].Data)
 	assert.Equal("{{ string .Secrets.symmetric_key_shared }}", testFiles["no_templates"].Data)
 
 	_, err = json.Marshal(testFiles)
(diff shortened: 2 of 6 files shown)


coordinator/core/marbleapi.go
```diff
@@ -279,20 +280,34 @@ func customizeParameters(params manifest.Parameters, specialSecrets reservedSecr
...
 
 	for name, data := range params.Env {
-		newValue, err := parseSecrets(data, secretsWrapped)
-		if err != nil {
-			return nil, err
+		if data.NoTemplates {
+			newValue = data.Data
+		} else {
+			if strings.Contains(data.Data, string([]byte{0x00})) {
+				return nil, fmt.Errorf("environment variable: %s: content contains null bytes", name)
```
Nirusu, line 305: Would it not be better if we straight out disallow pumping `raw` to `Env`? It would have failed before too during `Setenv`. While this error message is more user-friendly than before, it is something that can fail occasionally after manifest deployment (e.g. non-shared secrets which are regenerated, user-defined secrets...).

test/manifests.go
```diff
@@ -56,12 +56,20 @@ const ManifestJSON string = `{
...
+						"Encoding": "base64",
+						"NoTemplates": true
+					}
 				},
 				"Env": {
 					"IS_FIRST": "true",
 					"SEAL_KEY": "{{ hex .Marblerun.SealKey }}",
-					"TEST_SECRET_SYMMETRIC_KEY": "{{ raw .Secrets.symmetric_key_shared }}",
+					"TEST_SECRET_SYMMETRIC_KEY": {
+						"Data": "{{ raw .Secrets.symmetric_key_shared }}",
```
Nirusu, line 70: Given that the code now explicitly checks against NULL bytes passed to the environment, is this supposed to fail? Actually not right now regarding the tests if we check this or we don't.

Nirusu (commented):
Some feedback regarding my favorite pain points.

Functionality wise I am pretty happy with this so far, have not run into any issues so far :)

coordinator/core/marbleapi.go
daniel-weisse: `raw` can be used for user-defined secrets, which might contain arbitrary data, including valid utf-8 strings.
Another thing is that most non utf-8 data is still valid data for env variables (except null bytes of course).
So completely disallowing users to use the `raw` keyword for env variables is not a good idea, in my opinion.

This is also not fixed by introducing a new keyword (e.g. utf-8) that only allows valid utf-8 content in env vars, since there might still be a case where a randomly generated key is valid utf-8 and therefore get accepted.
I think it would be best to leave the option, but not advertise it by removing these cases from examples.

test/manifests.go
daniel-weisse: This should be replaced with a different key, probably `base64` or `hex`

coordinator/manifest/manifest_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+func TestFile(t *testing.T) {
+	dataJSON := []byte(`
+{
+	"string": "helloworld",
+	"string_struct": {
```
m1ghtym0, line 26:
```suggestion
	"stringStruct": {
```

coordinator/manifest/manifest_test.go
```diff
@@ -0,0 +1,97 @@
...
+	},
+	"base64_value": {
+		"Encoding": "string",
+		"Data": "YmFy"
+	},
+	"hex": {
+		"Encoding": "hex",
+		"Data": "4d6172626c6552756e"
+	},
+	"no_templates": {
```
m1ghtym0, line 44:
```suggestion
	"noTemplates": {
```

samples/sample-manifest.json
```diff
@@ -73,6 +92,18 @@
...
     "Secrets": {
+        "hello": {
+            "Type": "plain",
+            "UserDefined": true
+        },
+        "world": {
+            "Type": "plain",
+            "UserDefined": true
+        },
+        "api_key": {
```
m1ghtym0, line 103:
```suggestion
        "apiKey": {
```

samples/sample-manifest.json
```diff
@@ -20,13 +20,32 @@
...
                 },
                 "Env": {
                     "IS_FIRST": "true",
+                    "IS_SECOND": {
+                        "Encoding": "string",
+                        "Data": "false"
+                    },
                     "SEAL_KEY": "{{ hex .Marblerun.SealKey }}",
-                    "TEST_SECRET_SYMMETRIC_KEY": "{{ raw .Secrets.symmetric_key_shared }}",
+                    "API_KEY": "{{ string .Secrets.api_key }}",
```
m1ghtym0, line 48:
```suggestion
                    "API_KEY": "{{ string .Secrets.apiKey }}",
```

coordinator/manifest/manifest_test.go
```diff
@@ -0,0 +1,97 @@
...
+	testFiles := make(map[string]File)
+	err := json.Unmarshal(dataJSON, &testFiles)
+	require.NoError(err)
+	assert.Equal("helloworld", testFiles["string"].Data)
+	assert.Equal("string", testFiles["string"].Encoding)
+	assert.Equal("foo", testFiles["string_struct"].Data)
+	assert.Equal("bar", testFiles["base64"].Data)
+	assert.Equal("YmFy", testFiles["base64_value"].Data)
+	assert.Equal("MarbleRun", testFiles["hex"].Data)
+	assert.Equal("{{ string .Secrets.symmetric_key_shared }}", testFiles["no_templates"].Data)
```
m1ghtym0, line 62:
```suggestion
	assert.Equal("{{ string .Secrets.symmetric_key_shared }}", testFiles["noTemplates"].Data)
```

coordinator/manifest/manifest_test.go
```diff
@@ -0,0 +1,97 @@
...
+}`)
+	assert := assert.New(t)
+	require := require.New(t)
+
+	testFiles := make(map[string]File)
+	err := json.Unmarshal(dataJSON, &testFiles)
+	require.NoError(err)
+	assert.Equal("helloworld", testFiles["string"].Data)
+	assert.Equal("string", testFiles["string"].Encoding)
+	assert.Equal("foo", testFiles["string_struct"].Data)
```
m1ghtym0, line 58:
```suggestion
	assert.Equal("foo", testFiles["stringStruct"].Data)
```

m1ghtym0 (changes requested):
We should remove the other snake cases in a separate PR

samples/sample-manifest.json
daniel-weisse: Just noticed that the original proposal included the renaming of `raw` to `string`.
Are we set on this? Or is this still open to discussion?
What are your thoughts on this @Nirusu ?

coordinator/core/marbleapi.go
Nirusu: The reason why I dislike the "support, but not advertise" approach very much in this case is that this allows for spontaneous crashes after deployment. If you define a shared secret, say this:

```json
        "random_crap": {
            "Size": 256,
            "Shared": false,
            "Type": "symmetric-key"
        }
```
and you define this as Env:
```json
                "Env": {
                    "DO_I_BREAK": "{{ raw .Secrets.random_crap }}"
                },
```

And you create new Marbles occasionally... this goes well for quite some time, but some will just crash badly and will repeat to do so.

Having runtime crashes is not a good thing to have, and breaks reliability. It's a game of luck in this case, **and the only way to get rid of this is to redeploy the Coordinator from scratch**, as you cannot update this part of the manifest. Essentially, this allows a Coordinator deployment that's broken and cannot be fixed.

This is very bad UX, in my opinion. A solution such as MarbleRun should always aim for reliability. Sure, we are not in production mode yet, but having crashes by luck _intentionally supported_ is not a good idea.

Of course one can say: "But environment variables can contain more than just UTF-8/Unicode!", well yes... But this would just limit passing raw secrets into the environment. If you manually like to define bytes, you can do so via JSON control characters. It's not like we would completely neuter environment variables, just not support passing potentially random data into it causing crashes by chance.

However, I do not see the use case where someone actually wants to pass bytes through the environment via Secrets. Secrets usually contain cryptographic material, and having cryptographic material with NULL bytes removed _by design_ already seems pretty bad to me. If you use user-defined secrets to pass credentials, well that's usually UTF-8.

I really do not know what the use-case here would be, but supporting this is a reliability issue.

daniel-weisse: I agree this can lead to some annoying situations, but fixing everything that is wrong with templates and env var in their current form goes beyond the scope of this PR.
We should take some time to discuss solutions to this problem and then implement a fix.

samples/sample-manifest.json
m1ghtym0: I think keeping `raw` for now is fine. There is too much confusion going on already.

m1ghtym0 (approved):

Nirusu (approved):

PR #238, opened 2021-08-18
Remove legacy stuff

Nirusu:
### Proposed changes
- Remove `Marblerun.SealKey`
- Remove `Infrastructures`
- Remove `Clients`

Ever since we have secrets, this UUID derived SealKey thing appears to be obsolete. We used it in a couple of samples for Graphene, but it appears that the first one actually never really worked with Graphene (or it does not work with newer versions) due to a mismatch in key length (32 bytes instead of 16 bytes). 

On the same page, I decided to remove `Infrastructures` and `Clients` from the manifest. We never implemented these, however we had the definitions flying around... In theory, if we ever decide to re implement this, we have the version history here. But for now, I think we can remove this.

### Additional info
- It would be great if someone can run again the integration test on an SGX machine. For some reason, mine failes at the `bad_marble` step for every release we have in our history... Properly something is borked here, and since I touched quote code  to remove the empty infrastructure check, a re-check by someone with a working setup would be appreciated!


daniel-weisse:
Don't know if we want to remove the never-implemented `Infrastructures` just yet.
Actual implementation of the feature is far down the roadmap, but I don't think needlessly removing them now is the best idea.

Nirusu:
If we actually plan to integrate this soon, then I can just drop the commit and restore the old state.

daniel-weisse:
SGX integration test works fine on my machine

m1ghtym0:
Giving it some thought, let's keep the Infrastructures and drop the commit.
On the same note, I'll put it on the Roadmap and create tickets.
I believe for production deployments this might be fairly relevant.

Nirusu:
Dropped commit.

m1ghtym0 (approved):
Thanks!

daniel-weisse (approved):

PR #239, opened 2021-08-18
AB#1176 Template sanity check

daniel-weisse:
### Proposed changes
- Perform a check on the templates of all file and env declarations during `cc.SetManifest`  to make sure:
  1. All Secrets that are to be injected actually exist
  2. Only valid encodings are used (For example: no symmetric key is assigned PEM encoding)
  3. No raw encoding for secrets injected as environment variables, except for secrets of type `plain`
      *  If a `plain` secret contains NULL bytes, the deployment of the Marble fails. Since `plain` secrets are user-defined, this can be fixed at run time.

### Additional info
I have also made some errors produce by invalid templates more clear. Instead of receiving `"invalid secret type"` when a template references a non existent secret, we now get `"secret does not exist"`.
Combined with the extra information of what Marble and what file/env contains the invalid declaration, this should help users to find their mistake more easily.

### DIFF
--- a/coordinator/manifest/manifest.go
+++ b/coordinator/manifest/manifest.go
@@ -461,18 +461,21 @@ func EncodeSecretDataToBase64(data interface{}) (string, error) {
 	return base64.StdEncoding.EncodeToString([]byte(raw)), nil
 }
 
-// EncodeEnvSecretDataToRaw encodes only secrets of type plain to raw byte string
-func EncodeEnvSecretDataToRaw(data interface{}) (string, error) {
+// EncodeSecretDataToString encodes secrets to C type strings (no NULL bytes allowed as part of the string)
+func EncodeSecretDataToString(data interface{}) (string, error) {
 	switch secret := data.(type) {
 	case Secret:
 		if secret.Type != "plain" {
-			return "", errors.New("only secrets of type plain are allowed to use raw encoding for environment variables")
+			return "", errors.New("only secrets of type plain are allowed to use string encoding for environment variables")
+		}
+		if strings.Contains(string(secret.Public), string([]byte{0x00})) {
+			return "", errors.New("secret contains null bytes")
 		}
 		return EncodeSecretDataToRaw(data)
 	case nil:
 		return "", errors.New("secret does not exist")
 	default:
-		return "", errors.New("only secrets of type plain are allowed to use raw encoding for environment variables")
+		return "", errors.New("only secrets of type plain are allowed to use string encoding for environment variables")
 	}
 }
 
@@ -488,7 +491,7 @@ var ManifestFileTemplateFuncMap = template.FuncMap{
 var ManifestEnvTemplateFuncMap = template.FuncMap{
 	"pem":    EncodeSecretDataToPem,
 	"hex":    EncodeSecretDataToHex,
-	"raw":    EncodeEnvSecretDataToRaw,
+	"string": EncodeSecretDataToString,
 	"base64": EncodeSecretDataToBase64,
 }
 
--- a/coordinator/manifest/manifest.go
+++ b/coordinator/manifest/manifest.go
@@ -404,8 +404,10 @@ func EncodeSecretDataToPem(data interface{}) (string, error) {
 		typ, bytes = "PUBLIC KEY", secret
 	case PrivateKey:
 		typ, bytes = "PRIVATE KEY", secret
+	case nil:
+		return "", errors.New("secret does not exist")
 	default:
-		return "", errors.New("invalid secret type")
+		return "", errors.New("invalid secret type for pem encoding")
 	}
 
 	if len(bytes) <= 0 {
@@ -438,6 +440,8 @@ func EncodeSecretDataToRaw(data interface{}) (string, error) {
 		raw = secret.Public
 	case Certificate:
 		raw = secret.Raw
+	case nil:
+		return "", errors.New("secret does not exist")
 	default:
 		return "", errors.New("invalid secret type")
 	}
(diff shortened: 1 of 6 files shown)


coordinator/manifest/manifest.go
```diff
@@ -457,14 +461,37 @@ func EncodeSecretDataToBase64(data interface{}) (string, error) {
...
 }
 
-// ManifestTemplateFuncMap defines the functions which can be specified for secrets in the in go template format
-var ManifestTemplateFuncMap = template.FuncMap{
+// EncodeEnvSecretDataToRaw encodes only secrets of type plain to raw byte string
+func EncodeEnvSecretDataToRaw(data interface{}) (string, error) {
+	switch secret := data.(type) {
+	case Secret:
+		if secret.Type != "plain" {
+			return "", errors.New("only secrets of type plain are allowed to use raw encoding for environment variables")
```
Nirusu, line 469: I would still check if these secrets do contain unexpected NULL bytes and also throw an error in this case.

coordinator/manifest/manifest.go
```diff
@@ -457,14 +461,37 @@ func EncodeSecretDataToBase64(data interface{}) (string, error) {
 	return base64.StdEncoding.EncodeToString([]byte(raw)), nil
 }
 
-// ManifestTemplateFuncMap defines the functions which can be specified for secrets in the in go template format
-var ManifestTemplateFuncMap = template.FuncMap{
+// EncodeEnvSecretDataToRaw encodes only secrets of type plain to raw byte string
+func EncodeEnvSecretDataToRaw(data interface{}) (string, error) {
```
Nirusu, line 465: I don't quite like to re-use the type `raw` here but impose limits onto it when we don't do it for Files. This might be unexpected.

Nirusu (commented):
Quick review, have not tried to test (or break) it yet.

coordinator/manifest/manifest.go
daniel-weisse: I have added a general check for NULL bytes during `SetManifest` and moved the check at runtime during a Marble activation so it catches all NULL bytes.

I think this is better than to have the check in the template function itself, since NULL bytes can also occur if we dont use templates at all. For example through `base64` encoded env variables.

Nirusu: I think checks in `SetManifest` and `WriteSecrets` for user-defined secrets are sufficient, does not necessarily be in the template function if these are the only two functions deploying new secrets.

An extra error during Marble activation is fine I guess, though not necessary cause NULL bytes are actually catched by `os.Setenv`. It's just the error is ugly and non-explanatory by default (it's just "invalid argument").

coordinator/manifest/manifest.go
m1ghtym0: What about a more explicit type like `env` or smth.?

coordinator/manifest/manifest.go
m1ghtym0: > I think checks in `SetManifest` and `WriteSecrets` for user-defined secrets are sufficient, does not necessarily be in the template function if these are the only two functions deploying new secrets.
> 
> An extra error during Marble activation is fine I guess, though not necessary cause NULL bytes are actually catched by `os.Setenv`. It's just the error is ugly and non-explanatory by default (it's just "invalid argument").

I don't quite understand what you're trying to say here, could you rephrase?:-)

Nirusu: Basically check for (potential) NULL bytes whenever:

- We set a new manifest
- Someone updates an user-defined secret (which is piped into the environment)

Which should disallow any NULL bytes ideally. If for some other reasons there are still NULL bytes inserted during Marble activation, it will fail anyway, so we do not need any additional checks during runtime in my opinion, except if we want to return a user-friendly error to the user.

coordinator/manifest/manifest.go
Nirusu: The problem with `env` is that when you can also inject stuff like base64, what is `env` supposed to represent? I would just go with `(c)string` maybe.

coordinator/manifest/manifest.go
m1ghtym0: Thanks! Agreed

coordinator/manifest/manifest.go
m1ghtym0: Ok, let's move on here. I'm fine with `string`, if Daniel agrees let's just use it:-)

daniel-weisse: string is OK.
As long as we make it clear in the documentation which tags do what, there shouldn't be any problems

coordinator/manifest/manifest.go
daniel-weisse: We now have checking for NULL bytes on manifest set, and every time a Marble activation occurs

For manifest set we do a general check on the env variable string to catch any NULL bytes explicitly written by the person setting the manifest (e.g. a base64 encoded string which contains NULL bytes).

For a Marble activation, this check occurs directly in the template function: If the encoding is of type `string` (formerly `raw`), only data without NULL bytes result in a starting Marble. Otherwise an error is returned notifying the user of the NULL byte.

Nirusu: Okay, not I get the runtime checks. But should we not add another check for `WriteSecrets` upon upload of an user-defined secret?

Nirusu: I would just re-run the template dry-run. If it's just defined as a file, it should be fine, if it's defined as env, it should fail. Similar to as when you would try to launch such a Marble.

daniel-weisse: Added a template dry run in `WriteSecrets` :)

Nirusu (approved):

PR #240, opened 2021-08-18
AB#1177: Replace `sgxsdk-package-info` with a generic PackageProperties determination command

Nirusu:
### Proposed changes
- Replaces `sgxsdk-package-info` with `package-info`
- Add support for Open Enclave / Edgeless RT / EGo, Graphene and Occlum binaries
- For the latter two, pointing to the directory should be sufficient

### Additional info
- In the last commit, I attempted to design the command in a way that makes specifying the exact type of instance redundant and instead just try every method until we do not get an error. The code flow is a bit wild, so I am open for any suggestions on making this a tad bit cleaner.


cli/cmd/packageInfo.go
```diff
@@ -0,0 +1,248 @@
...
+	// Display the determined properties
+	if isOcclumInstance {
+		color.Cyan("PackageProperties for Occlum image at '%s':\n", path)
+	} else {
+		color.Cyan("PackageProperties for '%s':\n", path)
+	}
+
+	printPackageProperties(mrenclave, mrsigner, isvprodid, isvsvn)
+
+	os.Exit(0)
```
thomasten, line 113:
```suggestion
```
I'd prefer to let the caller check the error instead. It's a bit more LOC, but the caller's control flow is currently unobvious.

cli/cmd/packageInfo.go
```diff
@@ -0,0 +1,248 @@
...
+	// If the path is a directory, we try to find out if it's an Occlum image directory
+	var elfFile *elf.File
+	var isOcclumInstance bool
+	var err error
+	if isDirectory {
+		if elfFile, err = elf.Open(filepath.Join(path, "build/lib/libocclum-libos.signed.so")); err == nil {
+			color.Green("Detected Occlum image.")
+			isOcclumInstance = true
+		} else if os.IsNotExist(err) {
+			return err
```
thomasten, line 77:
```suggestion
```

cli/cmd/packageInfo.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil, nil, nil, nil, errors.New("could not find SIGSTRUCT header in given file")
+	}
+
+	// The Intel Software Developer Manual specifies SIGSTRUCT entries up to 1808 bytes.
+	// We use this as a cutoff for our sigStruct slice.
+	if len(sgxMetaData) <= sigStructIndex+1807 {
+		return nil, nil, nil, nil, errors.New("SGX metadata/SIGSTRUCT appears to be too small")
+	}
+
+	sigStruct := sgxMetaData[sigStructIndex : sigStructIndex+1807]
```
thomasten, line 136:
```suggestion
	const sigStructLen = 1808
	if len(sgxMetaData) < sigStructIndex+sigStructLen {
		return nil, nil, nil, nil, errors.New("SGX metadata/SIGSTRUCT appears to be too small")
	}

	sigStruct := sgxMetaData[sigStructIndex : sigStructIndex+sigStructLen]
```
I think the error was only in the comparison because the slicing excludes the last element.

cli/cmd/packageInfo.go
```diff
@@ -0,0 +1,248 @@
...
+	// Check if directory contains a file ending in .sig
+	var sigFile string
+	if isDirectory {
+		fsInfo, err := ioutil.ReadDir(path)
+		if err != nil {
+			return err
+		}
+		foundSigFile := false
+		for _, entry := range fsInfo {
+			if strings.HasSuffix(entry.Name(), ".sig") {
```
thomasten, line 169:
```suggestion
			if filepath.Ext(entry.Name()) == ".sig" {
```

thomasten (changes requested):

thomasten (approved):

PR #253, opened 2021-09-16
AB#1173 Get trusted time

katexochen:
### Proposed changes
- Add trusted time (ttime) package that uses Roughtime to get a trusted time.
- Add Roughtime server list to the manifest.
- Let the Coordinator use trusted time as soon as a manifest with Roughtime servers is set.

Regarding the last point, I'm currently not sure if I got every relevant call to the time package replaced with the trusted time. There are also some calls where I'm not sure if it's worth or needed to replace them.

### DIFF
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -189,10 +189,15 @@ func NewCore(dnsNames []string, qv quote.Validator, qi quote.Issuer, sealer seal
 		return nil, err
 	}
 
-	// TODO(katexochen): not sure if this is the right place for this
 	servers, err := c.data.getTimeServers()
-	if err == nil {
+	if err != nil {
+		c.zaplogger.Warn("No trusted time source found. Proceeding with untrusted host time.")
+	} else {
 		c.time = ttime.NewTime(servers, c.zaplogger)
+		// set time of sealer, if sealer is a TimeUser
+		if _, ok := sealer.(ttime.TimeUser); ok {
+			sealer.(ttime.TimeUser).SetTime(c.time)
+		}
 	}
 
 	rootCert, err := c.data.getCertificate(sKCoordinatorRootCert)
--- a/coordinator/core/core.go
+++ b/coordinator/core/core.go
@@ -24,7 +24,6 @@ import (
 	"math"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/edgelesssys/marblerun/coordinator/manifest"
 	"github.com/edgelesssys/marblerun/coordinator/quote"
@@ -32,6 +31,7 @@ import (
 	"github.com/edgelesssys/marblerun/coordinator/rpc"
 	"github.com/edgelesssys/marblerun/coordinator/seal"
 	"github.com/edgelesssys/marblerun/coordinator/store"
+	"github.com/edgelesssys/marblerun/coordinator/ttime"
 	"github.com/edgelesssys/marblerun/coordinator/updatelog"
 	"github.com/edgelesssys/marblerun/coordinator/user"
 	"github.com/edgelesssys/marblerun/util"
@@ -55,6 +55,7 @@ type Core struct {
 	updateLogger *updatelog.Logger
 	zaplogger    *zap.Logger
 	metrics      *coreMetrics
+	time         ttime.Time
 	rpc.UnimplementedMarbleServer
 }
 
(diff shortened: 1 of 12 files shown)


coordinator/ttime/ttime_test.go
```diff
@@ -0,0 +1,132 @@
...
+	// Create trusted time client.
+	tt := TrustedTime{servers: configs}
+
+	// Obtain time.
+	rt, err := tt.Roughtime()
+	require.NoError(err)
+
+	// Compare times. Obviously, this is a bidirectional test that heavily
+	// depends on the correctness of testing hardware's local time.
+	// TODO(katexochen): Maybe remove this later.
```
m1ghtym0, line 44: Ideally, this would not be part of a Unit test.
Maybe an integration test, but better our e2e-test.

m1ghtym0 (changes requested):

coordinator/core/core.go
```diff
@@ -187,6 +189,12 @@ func NewCore(dnsNames []string, qv quote.Validator, qi quote.Issuer, sealer seal
 		return nil, err
 	}
 
+	// TODO(katexochen): not sure if this is the right place for this
+	servers, err := c.data.getTimeServers()
```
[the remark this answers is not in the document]
katexochen: Yes, the idea was to only set a trusted time if the state contains a trusted time server list. Previous in the core generation, the time is set to an untrusted time.
We could check whether a manifest was set at this point, but since trusted time is optional in the manifest, this wouldn't give any additional information to our expectation to find a list of trusted time servers in the store.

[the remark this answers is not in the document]
m1ghtym0: I see, fair enough. Please add a log warning if we're running with untrusted time. Thanks!

coordinator/seal/seal.go
```diff
@@ -186,7 +186,7 @@ func (s *AESGCMSealer) SetEncryptionKey(encryptionKey []byte) error {
 // backupEncryptionKey creates a backup of an existing seal key
 func (s *AESGCMSealer) backupEncryptionKey() {
 	if sealedKeyData, err := ioutil.ReadFile(s.getFname(SealedKeyFname)); err == nil {
-		t := time.Now()
+		t := time.Now() // TODO(katexochen): trusted time?
```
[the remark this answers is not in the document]
katexochen: I've added trusted time to the `AESGCMSealer`. However, this won't prevent the named attack, since an malicious host could just block any connection to trusted time servers, and `ttime.Now()` would always return the same default `time.Time{}`.

m1ghtym0 (approved):
Thanks! Please add this to the docs as well:-)

coordinator/ttime/ttime.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"time"
+
+	"github.com/cloudflare/roughtime"
+	"github.com/cloudflare/roughtime/config"
+	"go.uber.org/zap"
+)
+
+// Time is an intrface that gives you the current time.
```
daniel-weisse, line 17:
```suggestion
// Time is an interface that gives you the current time.
```

coordinator/manifest/manifest.go
```diff
@@ -338,6 +342,26 @@ func (m Manifest) Check(ctx context.Context, zaplogger *zap.Logger) error {
...
+		}
+		seenNames[srv.Name] = true
+
+		if srv.PublicKeyType != "ed25519" {
+			return fmt.Errorf("invalid public key type %q specified for Roughtime server %q", srv.PublicKeyType, srv.Name)
+		}
+
+		udpAddr, err := serverUDPAddr(&srv)
+		if err != nil {
+			return fmt.Errorf("Roughtime server %q lists invalid address: %s", srv.Name, err)
```
daniel-weisse, line 358:
```suggestion
			return fmt.Errorf("roughtime server %q lists invalid address: %s", srv.Name, err)
```
error strings should not be capitalized

coordinator/ttime/ttime.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// UntrustedTime is just a wrapper around the default time package.
+// The time package can't be trusted since it uses the host's time.
+type UntrustedTime struct{}
+
+// Now is a wrapper around time.Now().
+func (u UntrustedTime) Now() time.Time {
+	return time.Now()
+}
```
daniel-weisse, line 75: Can we use a mockable time here?
Something like https://github.com/benbjohnson/clock would allow us to easily mock host time for roughtime testing purposes

coordinator/manifest/manifest.go
katexochen: Roughtime is a proper noun, which is[ capitalized even in error strings](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

PR #254, opened 2021-09-16
Rename helm chart to marblerun

m1ghtym0:
### Proposed changes
- Rename chart to marblerun
- Adjust CLI
- Use simulation as bool

### DIFF
--- a/cli/cmd/util.go
+++ b/cli/cmd/util.go
@@ -25,7 +25,7 @@ const webhookName = "marble-injector.marblerun"
 const (
 	helmChartName             = "edgeless/marblerun"
 	helmCoordinatorDeployment = "marblerun-coordinator"
-	helmInjectorDeployment    = "marbler-injector"
+	helmInjectorDeployment    = "marble-injector"
 	helmNamespace             = "marblerun"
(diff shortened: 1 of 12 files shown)


daniel-weisse (changes requested):
We also need a change for the [uninstall command](https://github.com/edgelesssys/marblerun/blob/7bd6a0cd6d749228724a863dfe66f52c2f514828/cli/cmd/uninstall.go#L68).

Maybe move the chart name to a global variable?

daniel-weisse (changes requested):

cli/cmd/util.go
```diff
@@ -21,6 +21,17 @@ import (
 
 const webhookName = "marble-injector.marblerun"
 
+// helm constants
+const (
+	helmChartName             = "edgeless/marblerun"
+	helmCoordinatorDeployment = "marblerun-coordinator"
+	helmInjectorDeployment    = "marbler-injector"
```
daniel-weisse, line 28:
```suggestion
	helmInjectorDeployment    = "marble-injector"
```

daniel-weisse (approved):

PR #257, opened 2021-09-17
Fix actions git config

m1ghtym0:
### Proposed changes
- See https://github.com/actions/checkout#push-a-commit-using-the-built-in-token
- Not sure what name, email is used by default and why this works sometimes...


katexochen (approved):
LGTM :+1:

PR #260, opened 2021-09-27
AB#1302 Remove namespace command

daniel-weisse:
### Proposed changes
- Remove the `namespace` command
- Trigger marble-injector webhook based on pod labels instead of namespace labels

For a Pod to be injected by MarbleRun, we require both a label on the Pod itself (`marblerun/marbletype`), and a label on the namespace it is starting in (`marblerun/inject=enabled`).
We can simplify the deployment process by instead injecting any Pod with the `marblerun/marbletype` label.

This removes the need for the namespace command.

To disable injection of sgx resources, users can add the `marblerun/resource-injection=disabled` label.

### Additional info
This will require updates to all demos and tests making use of the `namespace` command as soon as we have a new MarbleRun release.


charts/templates/marble-injector.yaml
```diff
@@ -99,31 +99,11 @@ webhooks:
       apiVersions: ["v1"]
       resources: ["pods"]
       scope: "Namespaced"
```
m1ghtym0, line 101: Is the scope still "namespaced"?

daniel-weisse: If I understand the kubernetes scope concept correctly a webhook with scope `"Namespaced"` is supposed to match resources that can be associated with namespaces (e.g. Pods). 
The two other options are `"Cluster"` for cluster-scoped resources (resources that do not belong to a namespace), and `""` which means there are no scope restrictions.

So `Namespaced` should still be correct

m1ghtym0: Thanks for clearing that up!

m1ghtym0 (approved):

Issue #290, opened 2022-03-14
Coordinator API - endpoint for manifest signature


lead4good:
### Use case
Attestation in air gapped environments is currently not possible if the coordinator API is not part of that environment. Without the secure and attested connection to the API it is not possible to attestate the manifest.

### Describe your solution
The coordinator attestation report, the PKI chain and the manifest can be distributed to an air gapped environment easily. What is missing is a manifest signature that links the PKI chain to the manifest and hence to the attestation report. The intermediate key should be used to sign the manifest to securely handle manifest updates.

### Additional context
For a real world use case imagine trying to attestate marbles from within a blockchain. This is currently not possible as the blockchain itself cannot access the coordinator API.

m1ghtym0:
Thank you @lead4good, this sounds useful!
Would you be interested to contribute the implementation?

lead4good:
@m1ghtym0 Could be. Let me check with the stakeholders what priority this feature has.

lead4good:
@m1ghtym0 I've received the OK from the stakeholders.

Let me describe the implementation I have in mind.
The coordinators `manifest/get` [endpoint](https://github.com/edgelesssys/marblerun/blob/590f82fcd4bd6885ecdd22037b243c82fdc992a1/coordinator/server/client_api.go#L104-L110) already returns the base64 encoded manifest plus a hash of the manifest. I would append this endpoint to also return another value, which will be the hashes signature - signed by the intermediate key.
```
rsa.SignPKCS1v15(rand.Reader, intermediate.PrivateKey, crypto.SHA256, manifest_hash)
## equivalent to
openssl dgst -sha256 -binary -sign intermediate.pem manifest.json | base64
```
I would add this logic in the `GetManifestSignature` [function](https://github.com/edgelesssys/marblerun/blob/590f82fcd4bd6885ecdd22037b243c82fdc992a1/coordinator/core/clientapi.go#L216-L223).

I am not yet sure how to name the signature parameter as a parameter `ManifestSignature` already exists (it returns the hash). To prevent confusion, the hash values parameter name should be `ManifestHash` or `ManifestDigest`, while the signatures parameter name should in fact be `ManifestSignature`. This would introduce breaking changes however.

The steps to verify the signature would then be
```
curl --cacert marblerun.crt "https://$MARBLERUN/manifest" | jq '.data.ManifestSignature' --raw-output | base64 -d > manifest.sig
[...]

PR #291, opened 2022-03-26
Coordinator API - endpoint for manifest signature

lead4good:
### Proposed changes
- enhance the Coordinator API `manifest` endpoint to return the manifests ECDSA signature - signed by intermediate key

### Related issue
- #290 

### Open Questions
- The ECDSA signature contains a random element, which means that with the current implementation the signature will change upon each request. An attacker who would like to guess the ECDSA key might find this useful. It could be sensible to create the signature only once and store it in the storeWrapper. @m1ghtym0 any preferences?

### DIFF
--- a/coordinator/core/clientapi_test.go
+++ b/coordinator/core/clientapi_test.go
@@ -44,9 +44,9 @@ func TestGetManifestSignature(t *testing.T) {
 	sigECDSA, hash, manifest := c.GetManifestSignature(context.TODO())
 	expectedHash := sha256.Sum256([]byte(test.ManifestJSON))
 	assert.Equal(expectedHash[:], hash)
-	intermediatePrivK, err := c.data.getPrivK(sKCoordinatorIntermediateKey)
+	rootPrivK, err := c.data.getPrivK(sKCoordinatorRootKey)
 	assert.Equal(err, nil)
-	assert.True(ecdsa.VerifyASN1(&intermediatePrivK.PublicKey, expectedHash[:], sigECDSA))
+	assert.True(ecdsa.VerifyASN1(&rootPrivK.PublicKey, expectedHash[:], sigECDSA))
 	assert.Equal([]byte(test.ManifestJSON), manifest)
 }
 
--- a/coordinator/core/clientapi_test.go
+++ b/coordinator/core/clientapi_test.go
@@ -45,7 +45,7 @@ func TestGetManifestSignature(t *testing.T) {
 	expectedHash := sha256.Sum256([]byte(test.ManifestJSON))
 	assert.Equal(expectedHash[:], hash)
 	rootPrivK, err := c.data.getPrivK(sKCoordinatorRootKey)
-	assert.Equal(err, nil)
+	assert.NoError(err)
 	assert.True(ecdsa.VerifyASN1(&rootPrivK.PublicKey, expectedHash[:], sigECDSA))
 	assert.Equal([]byte(test.ManifestJSON), manifest)
 }
--- a/coordinator/core/clientapi_test.go
+++ b/coordinator/core/clientapi_test.go
@@ -8,6 +8,7 @@ package core
 
 import (
 	"context"
+	"crypto/ecdsa"
 	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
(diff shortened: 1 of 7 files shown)


lead4good:
[rfc6979](https://datatracker.ietf.org/doc/html/rfc6979) would also be an option to make the signature output deterministic, however, I can not find an implementation in go which is trustworthy.

daniel-weisse:
Writing the signature to the store seems like the most sensible idea for now.
I would add the generation and writing to store in [`SetManifest()`](https://github.com/edgelesssys/marblerun/blob/590f82fcd4bd6885ecdd22037b243c82fdc992a1/coordinator/core/clientapi.go#L46).

lead4good:
@daniel-weisse the manifest signature is now deterministic by setting it once and moving it to the store.
I've just now figured out how manifest updates are handled. As the original manifest is never touched - only update logs exist - it doesn't really make sense to sign the manifest with the intermediate key.
I would add another patch that uses the root key for signing the manifest. This would mean manifest updates still can not be verified by key signatures, but I would propose to postpone this issue to another PR.

lead4good:
@daniel-weisse Please review

coordinator/server/client_api.go
```diff
@@ -86,26 +89,43 @@ func (s *clientAPIServer) statusGet(w http.ResponseWriter, r *http.Request) {
...
 //		 500: ErrorResponse
 func (s *clientAPIServer) manifestGet(w http.ResponseWriter, r *http.Request) {
-	signature, manifest := s.cc.GetManifestSignature(r.Context())
+	signatureRootECDSA, signature, manifest := s.cc.GetManifestSignature(r.Context())
 	writeJSON(w, ManifestSignatureResp{
-		ManifestSignature: hex.EncodeToString(signature),
-		Manifest:          manifest,
+		ManifestSignatureRootECDSA: signatureRootECDSA,
+		ManifestSignature:                  hex.EncodeToString(signature),
+		Manifest:                           manifest,
```
daniel-weisse, line 128:
```suggestion
		ManifestSignatureRootECDSA: signatureRootECDSA,
		ManifestSignature:          hex.EncodeToString(signature),
		Manifest:                   manifest,
```

coordinator/server/client_api.go
```diff
@@ -39,6 +39,9 @@ type StatusResp struct {
 }
 
 type ManifestSignatureResp struct {
+	// The manifest signature - signed by the root ECDSA key. Changes upon each call due to a random component.
+	// example: MEYCIQCmkqOP0Jf1v5ZR0vUYNnMxmy8j9aYR3Zdemuz8EXNQ4gIhAMk6MCg00Rowilui/66tHrkETMmkPmOktMKXQqv6NmnN
+	ManifestSignatureRootECDSA []byte
```
daniel-weisse, line 44: Is it possible to have this as a hex encoded string, similar to the `ManifestSignature` field?
Or would that negatively impact the verification workflow?

coordinator/core/clientapi_test.go
```diff
@@ -40,9 +41,12 @@ func TestGetManifestSignature(t *testing.T) {
 
 	assert.NoError(err)
 
-	sig, manifest := c.GetManifestSignature(context.TODO())
+	sigECDSA, hash, manifest := c.GetManifestSignature(context.TODO())
 	expectedHash := sha256.Sum256([]byte(test.ManifestJSON))
-	assert.Equal(expectedHash[:], sig)
+	assert.Equal(expectedHash[:], hash)
+	rootPrivK, err := c.data.getPrivK(sKCoordinatorRootKey)
+	assert.Equal(err, nil)
```
daniel-weisse, line 48:
```suggestion
	assert.NoError(err)
```

daniel-weisse (changes requested):
Thanks for the PR.
Looks good to me over all, just three small things:

coordinator/server/client_api.go
lead4good: I personally prefer and would recommend to stick with base64. While it is common to display a hash hex encoded and one might compare hashes just by looking at them, you can not verify a signature like that. Hence, base64 decoding the signature and piping it to a file sounds much more simpler and securer than having to juggle around with tools such as sed, awk or xxd. Also there is potential for confusion (big endian, little endian).

coordinator/server/client_api.go
```diff
@@ -39,6 +39,9 @@ type StatusResp struct {
 }
 
 type ManifestSignatureResp struct {
+	// The manifest signature - signed by the root ECDSA key. Changes upon each call due to a random component.
+	// example: MEYCIQCmkqOP0Jf1v5ZR0vUYNnMxmy8j9aYR3Zdemuz8EXNQ4gIhAMk6MCg00Rowilui/66tHrkETMmkPmOktMKXQqv6NmnN
```
daniel-weisse, line 43:
```suggestion
	// example: MEYCIQCmkqOP0Jf1v5ZR0vUYNnMxmy8j9aYR3Zdemuz8EXNQ4gIhAMk6MCg00Rowilui/66tHrkETMmkPmOktMKXQqv6NmnN
	// swagger:strfmt byte
```
OpenAPI workflow does not work for pull request from forks.

You can test it locally using the following:
```shell
swagger generate spec -m --compact --exclude-deps -o ./swagger.json
swagger validate ./swagger.json
```

coordinator/server/client_api.go
```diff
@@ -39,6 +39,9 @@ type StatusResp struct {
 }
 
 type ManifestSignatureResp struct {
+	// The manifest signature - signed by the root ECDSA key. Changes upon each call due to a random component.
```
daniel-weisse, line 42:
```suggestion
	// The manifest signature - signed by the root ECDSA key.
```
Should stay consistent now

daniel-weisse (approved):

m1ghtym0 (approved):

PR #294, opened 2022-04-19
Remove nodename check for Occlum premain

Nirusu:
### Proposed changes

I have the hunch that the premain might not work for the most recent Occlum 0.27.1 as they [do now fill the `nodename` field for their `uname` struct](https://github.com/occlum/occlum/commit/ffdd4d95a453f6db226daa58d71d212392ae5556) with the hostname of the host machine. So let's just remove that check, the system name should be pretty evident already.

### Additional info
Untested. Actually, I have not tested if it actually breaks since I only stumbled upon this by accident while scrolling around GitHub... but I suspect it does.

### DIFF
--- a/cmd/premain-libos/main.go
+++ b/cmd/premain-libos/main.go
@@ -81,7 +81,7 @@ func detectLibOS() (int, error) {
 
 	// Occlum detection
 	// Taken from: https://github.com/occlum/occlum/blob/master/src/libos/src/misc/uname.rs
-	if sysname == "Occlum" && nodename == "occlum-node" {
+	if sysname == "Occlum" {
 		return occlum, nil
 	}
 


Nirusu:
Guess the GitHub Actions needs some adjustsments due to: https://github.blog/2022-04-12-git-security-vulnerability-announced/

(see https://github.com/actions/checkout/issues/760)

.github/workflows/unittests.yml
```diff
@@ -7,19 +7,21 @@ on:
...
+        uses: actions/checkout@v3
 
       - name: Test
         run: ertgo test -race -count=3 ./...
 
       - name: Setup
-        run: mkdir build
+        run: | 
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+          mkdir build
```
daniel-weisse, line 24:
```suggestion
        run: mkdir build
```
As far as I can tell the unit test issue was caused by an incompatibility with old git versions and Go 1.18 (we recently updated the used Go version in our CI image), so this should not be needed here

daniel-weisse (commented):
Tested this and our current premain does indeed not work with the latest Occlum version. The update in this PR fixes that. Thanks!

Also thanks for trying to fix the CI, sadly our openAPI workflow has some issues with external contributions. However, unit tests work again :smile: 

For our Occlum sample to work with the latest version, it requires some small changes.
Could you add them to this PR?

In `samples/occlum-hello/Dockerfile`:
```diff
- FROM occlum/occlum:0.24.1-ubuntu18.04
+ FROM occlum/occlum:0.27.1-ubuntu20.04
```

In `samples/occlum-hello/Occlum.json`:
```diff
   "metadata": {
        "product_id": 0,
        "version_number": 0,
        "debuggable": true,
+        "enable_kss": false,
+        "family_id": {
+            "high": "0x0",
+            "low": "0x0"
+        },
+        "ext_prod_id": {
+            "high": "0x0",
+            "low": "0x0"
+        }
    },
    "mount": [
```

.github/workflows/unittests.yml
Nirusu: Does not appear to be fixed with Go 1.18.

See: 
https://github.com/edgelesssys/marblerun/runs/6213150125 (works with `git config --global --add safe.directory "$GITHUB_WORKSPACE"`)
https://github.com/edgelesssys/marblerun/runs/6213170315 (does not work without it)

samples/occlum-hello/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Replace standard Intel DCAP plugin with Azure DCAP plugin
-FROM occlum/occlum:0.24.1-ubuntu18.04
+FROM occlum/occlum:0.27.1-ubuntu18.04
```
daniel-weisse, line 2:
```suggestion
FROM occlum/occlum:0.27.1-ubuntu20.04
```

daniel-weisse (changes requested):
Works fine if we build from the correct image :D

Nirusu:
Oops, fixed.

PR #305, opened 2022-07-13
allow setting a manifest from env on startup

### LINKED ISSUE #303: Take manifest from cmdline
Client is requesting that they went to start marblerun + edb from a single docker compose file.

We will implement this feature but would it be ok to submit it upstream? 

I think it would probably just involve adding a cli flag that takes an initial manifest.

aep:
fixes #303

### DIFF
--- a/cmd/coordinator/run.go
+++ b/cmd/coordinator/run.go
@@ -7,11 +7,11 @@
 package main
 
 import (
+	"context"
+	"io/ioutil"
 	"log"
 	"os"
 	"strings"
-	"io/ioutil"
-	"context"
 
 	"github.com/edgelesssys/marblerun/coordinator/config"
 	"github.com/edgelesssys/marblerun/coordinator/core"
--- a/cmd/coordinator/run.go
+++ b/cmd/coordinator/run.go
@@ -10,6 +10,8 @@ import (
 	"log"
 	"os"
 	"strings"
+	"io/ioutil"
+	"context"
 
 	"github.com/edgelesssys/marblerun/coordinator/config"
 	"github.com/edgelesssys/marblerun/coordinator/core"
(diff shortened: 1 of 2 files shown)


cmd/coordinator/run.go
```diff
@@ -10,6 +10,8 @@ import (
 	"log"
 	"os"
 	"strings"
+	"io/ioutil"
+	"context"
```
daniel-weisse, line 14:
```suggestion
	"context"
	"io/ioutil"
	"log"
	"os"
	"strings"
```
Just a quick `gofmt`, otherwise looks good

aep:
> Just want to point out that this follows the usual MarbleRun trust model:

i think you mean _doesnt_ ?

yeah, the intended use case appears to be to run the coordinator in a "trusted" environment, i.e. outside of the cloud provider.
on the other hand, would it make the feature more generally useful if we would embed the manifest into the binary so it is signed?

m1ghtym0:
> > Just want to point out that this follows the usual MarbleRun trust model:
> 
> i think you mean _doesnt_ ?
> 
No, I actually meant _does_.
That is generally how MarbleRun works.
Only after setting a manifest, MarbleRun is considered initialized and can be verified by remote attestation.
Just wanted to point that out, since, there might be confusion about the trust model here when the manifest comes from an untrusted environment.

> yeah, the intended use case appears to be to run the coordinator in a "trusted" environment, i.e. outside of the cloud provider. on the other hand, would it make the feature more generally useful if we would embed the manifest into the binary so it is signed?
> 
> that would unfortunately require recompiling it every time the manifest changes.

No need for that. The MarbleRun attestation process will verify that the deployed manifest is correct and wasn't modified.

daniel-weisse (approved):

PR #308, opened 2022-07-21
Fix an error that occurs when get a coordinator certificate

jcsora:
### Proposed changes
- use "era-config.json" instead of configFilename variable when the latter is empty.

### DIFF
--- a/cli/cmd/util.go
+++ b/cli/cmd/util.go
@@ -34,8 +34,7 @@ const (
 
 const promptForChanges = "Do you want to automatically apply the suggested changes [y/n]? "
 
-const ERADefaultConfig = "era-config.json"
-
+const eraDefaultConfig = "era-config.json"
 var (
 	eraConfig   string
 	insecureEra bool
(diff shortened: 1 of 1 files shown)


m1ghtym0:
Fix LGTM, not sure if we want to move that string into a constant?

jcsora:
> Fix LGTM, not sure if we want to move that string into a constant?

I updated the code, can you please review it?

jcsora:
> @jcsora Can you sign the CLA, please?

Yes i just finished signing.

cli/cmd/util.go
```diff
@@ -34,6 +34,8 @@ const (
 
 const promptForChanges = "Do you want to automatically apply the suggested changes [y/n]? "
 
+const ERADefaultConfig = "era-config.json"
+
```
daniel-weisse, line 38:
```suggestion
const eraDefaultConfig = "era-config.json"
```
Filename does not need to be exported, otherwise looks good to me, thanks

jcsora:
> @jcsora Can you sign the CLA, please?

Yes, I can sign, but I don't have much experience with this, how can I do this?

daniel-weisse (approved):

Issue #307, opened 2022-07-20
emit cert as Java KeyStore 


aep:
### Use case

java requires using a "keystore" instead of just pem encoded certs.
so we'll have to emit those from premain.

### Describe your solution

(https://github.com/pavlo-v-chernykh/keystore-go) could be used to provide the nessesary template func

something like

```
"Marbles": {
        "javathing": {
            "Parameters": {
                "Files": {
                    "/app/keystore.jks": "{{ javaKeyStore .MarbleRun.MarbleCert.Cert }}",
```

daniel-weisse:
I would prefer not to add language specific encodings to the manifest.
In my opinion, parsing files is something that should be left to the application itself.

If PEM is something your application does not handle well, you can also use `raw`, to get the certificate in DER encoded certificate, or `hex`, to get the DER certificate hex encoded. See [the docs](https://docs.edgeless.systems/marblerun/#/workflows/define-manifest?id=templates) for more details.

Tagging @m1ghtym0 in case he has a different opinion.

m1ghtym0:
I agree with Daniel that ideally, encodings should be language neutral.
However, I see the problem here and it's just another encoding.
IMO, that's not a big deal and we can have as many encodings as necessary, especially in hindsight of lift&shift.
So, @aep please feel free to add JavaKeyStore as a file encoding.

PR #309, opened 2022-07-21
RFC: emit cert as Java KeyStore

aep:
see #307

turns out jks is slightly more annoying because it is usually used in a multi-step process of adding certs into a store,
but we need to emit the file in one go, merging all the inputs.

so the proposed semantic is

```json
"/app/keystore.jks": "{{ jks 'password' 'alias'  .MarbleRun.MarbleCert.Private .MarbleRun.MarbleCert.Cert }}", 
"/app/trustme.jks": "{{ jks 'password' 'alias'  nil .MarbleRun.MarbleCert.Cert }}", 
```

we could add another helper function to make the structure more explicit, but i feel like it's a can of worms and this works fine for all sensible use cases.

we could also go the opposite way and reduce it to only accept manifest.Secret

```json
"/app/keystore.jks": "{{ jks 'password' 'alias' MarbleRun.MarbleCert 'alias2' MarbleRun.OtherCert  }}", 
```

which is alot prettier, but i don't know if people maybe have weird use cases where they want control over what exactly gets emitted.

on second thought, i'd much prefer that if we can get away with it.
So do you see any use cases where someone would want to emit something that is NOT a manifest.Secret?

### DIFF
--- a/coordinator/manifest/manifest.go
+++ b/coordinator/manifest/manifest.go
@@ -7,6 +7,7 @@
 package manifest
 
 import (
+	"bytes"
 	"context"
 	"crypto/x509"
 	"encoding/base64"
@@ -15,8 +16,10 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"github.com/pavlo-v-chernykh/keystore-go/v4"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/edgelesssys/marblerun/coordinator/quote"
 	"github.com/edgelesssys/marblerun/coordinator/user"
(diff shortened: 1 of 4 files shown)


m1ghtym0:
Let me clarify because I'm not that familiar with JKS.
The JKS only holds keys or certificates or can it store arbitrary binary blobs?
In case we only speak about keys&certs, IMO we can assume that these can always be represented using `manifest.Secret`.
That can include constants or Coordinator generated values.

aep:
> The JKS only holds keys or certificates or can it store arbitrary binary blobs?

The only relevant use case is to store certificates to use as trust root, and for storing private keys for a cert.

> In case we only speak about keys&certs, IMO we can assume that these can always be represented using manifest.Secret.

yeah i agree. i'm going to make it do that. i was just wondering if there's use cases in marblerun that i'm not seing. but since you can load pretty much everything into a manifest.Secret it should be fine.

.dockerignore
```diff
@@ -1 +1,2 @@
 build/
+Dockerfile
```
daniel-weisse, line 2:
```suggestion
```

coordinator/manifest/manifest.go
```diff
@@ -477,12 +480,64 @@ func EncodeSecretDataToString(data interface{}) (string, error) {
 	}
 }
 
+// EncodeSecretsToJavaKeyStore encodes secrets as java keystore
+func EncodeSecretsToJavaKeyStore(password string, secrets...Secret) (string, error) {
+	var ks = keystore.New()
```
daniel-weisse, line 485:
```suggestion
	ks := keystore.New()
```

coordinator/manifest/manifest.go
```diff
@@ -477,12 +480,64 @@ func EncodeSecretDataToString(data interface{}) (string, error) {
 	}
 }
 
+// EncodeSecretsToJavaKeyStore encodes secrets as java keystore
+func EncodeSecretsToJavaKeyStore(password string, secrets...Secret) (string, error) {
```
daniel-weisse, line 484:
```suggestion
// EncodeSecretsToJavaKeyStore encodes secrets as java keystore.
func EncodeSecretsToJavaKeyStore(password string, secrets ...Secret) (string, error) {
```

daniel-weisse (changes requested):
Thanks, looks good to me.
Just a couple of style things

aep:
done. let me know about additional changes.

the test only checks if the semantics work, it doesnt actually test if the keystore is valid.
i feel like that would drag in too much java stuff.

aep:
uuuh it turns out there might be a use case where you dont want to important the entire secret

specifically edb.

the service that connects to edb should NOT have the secret

m1ghtym0:
Hey @aep,

> I like the latter option better. You can pass `edbSecret.Cert` to the client Marble and `edbSecret` to edb. The handler function just handles both cases.

Does that sound good to you?

aep:
oh hey sorry about this. my customer hasn't really focused on this.
i think the best course of action is to keep the patch downstream.

PR #317, opened 2022-10-25
Fix dcap-qpl values reference

daniel-weisse:
### Proposed changes
- Update helm chart to allow the CLI to correctly parse the `dcap-qpl` flag

### Additional info
- We might want to update our existing helm charts for v0.5-v0.6, since they are broken without this change

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -33,7 +33,6 @@ their default values.
 
 | Parameter                                    | Type           | Description    | Default                              |
 |:---------------------------------------------|:---------------|:---------------|:-------------------------------------|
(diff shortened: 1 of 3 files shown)


thomasten:
Can you also adapt charts/README.md?

thomasten (approved):
> We might want to update our existing helm charts for v0.5-v0.6, since they are broken without this change

Alternative would be to treat this as a bug in those releases and create a new release?

Maybe we should make a new MarbleRun release anyway because PCCS API v3.0 is broken in the ERT release that was used for the last MarbleRun release.

PR #321, opened 2022-11-02
feat: dependabot workflow automation for updating dependency

Rajpratik71:
### Proposed changes
- Dependabot workflow automation for updating dependency

Signed-off-by: Pratik Raj <Rajpratik71@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
+    directory: "/" # Location of package manifests
+    schedule:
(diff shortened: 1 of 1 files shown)


m1ghtym0:
Hi @Rajpratik71, thanks for the PR!
We want to add e2e tests for MarbleRun soon, so we can automate the testing of changes, such as the ones introduced by dependabot for example. Right now we would have to carefully check the changes proposed by dependabot manually running the integration tests.
For security-relevant updates, we already get the dependabot notifications and PRs.
After we automated e2e testing we'll also add dependency tracking/management via renovate, similar to how we did with [Constellation](https://github.com/edgelesssys/constellation/blob/main/renovate.json).

PR #328, opened 2022-12-12
Separate client API from core package

daniel-weisse:
### Proposed changes

Refactor the monolithic `core` package to allow for a more modular build process.

To start of, this PR removes the client API part from the Core and lays the groundwork for future refactoring.
This is done through the following changes:

- Move the Client API from `core` to its own package (`clientapi`)
  -  No dependency on the `core` package
  - Only holds interfaces (except for logging)
  - `clientapi` implements the interface required by the `server` package
- Move `storewrapper` to its own package in `coordinator/store/wrapper` 
- Move `stdstore` to its own package in `coordinator/store/stdstore`
- Create `constants` package to consolidate old `config` package with constants that were defined in `core`


coordinator/core/core.go
```diff
@@ -101,56 +82,54 @@ func (c *Core) requireState(states ...state) error {
...
-	if !(curState < newState && newState < stateMax) {
+	if !(curState < newState && newState < state.Max) {
 		panic(fmt.Errorf("cannot advance from %d to %d", curState, newState))
 	}
-	return txdata.putState(newState)
+	return txdata.PutState(newState)
+}
+
+// Unlock the Core's mutex.
+func (c *Core) Unlock() {
```
thomasten, line 99: this is a bit shady, but we should keep it for now

coordinator/core/core.go
```diff
@@ -327,20 +306,31 @@ func generateCert(dnsNames []string, commonName string, privk *ecdsa.PrivateKey,
 	return cert, privk, nil
```
thomasten, line 306: can we remove this func (generateCert) and use the one from the crypto package?

coordinator/core/core.go
```diff
@@ -606,34 +596,34 @@ func generateUsersFromManifest(rawUsers map[string]manifest.User, roles map[stri
 }
```
thomasten, line 596: remove func generateUsersFromManifest

coordinator/core/core_test.go
```diff
@@ -73,42 +70,46 @@ func TestSeal(t *testing.T) {
...
-	cSecrets, err := c.data.getSecretMap()
+	cSecrets, err := c.data.GetSecretMap()
 	assert.NoError(err)
 
 	// Check sealing with a new core initialized with the sealed state.
-	c2, err := NewCore([]string{"localhost"}, validator, issuer, sealer, recovery, zapLogger, nil, nil)
+	c2, err := NewCore([]string{"localhost"}, validator, issuer, stdstore.New(sealer), recovery, zapLogger, nil, nil)
+	clientAPI, err = clientapi.New(c2.store, c2.recovery, c2, zapLogger)
+	require.NoError(err)
 	require.NoError(err)
```
thomasten, line 95:
```suggestion
	c2, err := NewCore([]string{"localhost"}, validator, issuer, stdstore.New(sealer), recovery, zapLogger, nil, nil)
	require.NoError(err)
	clientAPI, err = clientapi.New(c2.store, c2.recovery, c2, zapLogger)
	require.NoError(err)
```

coordinator/manifest/manifest.go
```diff
@@ -23,6 +23,13 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	SecretTypeCertECDSA    = "cert-ecdsa"
+	SecretTypeED25519      = "cert-ed25519"
+	SecretTypeCertRSA      = "cert-rsa"
+	SecretTypeSymmetricKey = "symmetric-key"
```
thomasten, line 30: can we use these at all places where the strings are still directly used?

thomasten (changes requested):

coordinator/core/core.go
daniel-weisse: I added it to not alter existing code too much.
This is basically the old `c.mux.Unlock` that was used by the client API.
Ideally I would like to refactor the `RequireState` logic to not require us to defer an unlock statement beforehand.
Maybe this could be down by letting `RequireState` return an unlock function?

thomasten: > I added it to not alter existing code too much

Yeah, I thought so.

Let's keep it for now and maybe add a TODO and/or create a ticket to do it after we finished the more important tasks.

thomasten (approved):

m1ghtym0 (approved):
Changes make sense and lgtm:-)

PR #329, opened 2022-12-14
Add logging for failed marble activations

daniel-weisse:
### Proposed changes
- Add logging in case of failed Marble activation

### DIFF
--- a/coordinator/core/marbleapi.go
+++ b/coordinator/core/marbleapi.go
@@ -56,7 +56,7 @@ type secretsWrapper struct {
 //
 // Returns a signed certificate-key-pair and the application's parameters if the authentication was successful.
 // Returns an error if the authentication failed.
-func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.ActivationResp, error) {
+func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (res *rpc.ActivationResp, err error) {
 	c.log.Info("Received activation request", zap.String("MarbleType", req.MarbleType))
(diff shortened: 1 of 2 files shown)


coordinator/core/marbleapi.go
```diff
@@ -87,12 +93,13 @@ func (c *Core) Activate(ctx context.Context, req *rpc.ActivationReq) (*rpc.Activ
 
 	marbleRootCert, err := c.data.GetCertificate(constants.SKMarbleRootCert)
 	if err != nil {
-		c.log.Error("Could not retrieve marbleRootCert certificate.", zap.Error(err))
+		c.log.Error("Couldn't retrieve marbleRootCert certificate", zap.Error(err))
```
m1ghtym0, line 96: Are we logging twice in these cases? Should we just wrap the error and return it, or do you think it makes sense to log in two steps?

daniel-weisse: Effectively yes.
Maybe we should do a quick pass over on the error return logic of marbleapi.go.
In many cases subfunctions are already returning gRPC errors which makes the whole thing a bit unwieldy.
The current implementation also logs the whole gRPC error, which might not be wanted.

Maybe deferring a `Marble activation failed` without showing the specific error, and providing more detailed error logging for the specific failures makes more sense.

For example, we would keep line 96 as it currently, but we won't print the error again in the `Marble activation failed` message.

m1ghtym0: I don't have a strong opinion as long as we make it consistent. I like clean and meaningful error message that dont' require digging through a ton of noise, so I like your proposal:-)

m1ghtym0 (approved):

PR #333, opened 2022-12-14
[premain-libos] remove nodename from gramine libos detection

lead4good:
In a non containerized setup the gramine libOS check fails due to the "localhost" nodename check.

### DIFF
--- a/cmd/premain-libos/main.go
+++ b/cmd/premain-libos/main.go
@@ -74,7 +74,6 @@ func detectLibOS() (int, error) {
 
 	// Clean utsname
 	sysname := strings.ReplaceAll(string(utsname.Sysname[:]), "\x00", "")
-	nodename := strings.ReplaceAll(string(utsname.Nodename[:]), "\x00", "")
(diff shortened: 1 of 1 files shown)


cmd/premain-libos/main.go
```diff
@@ -88,7 +87,7 @@ func detectLibOS() (int, error) {
 	// Gramine detection
 	// This looks like a general Linux kernel name, making it harder to detect... But it's unlikely someone is running SGX code on Linux 3.10.0.
 	// Taken from: https://github.com/gramineproject/gramine/blob/master/LibOS/shim/src/sys/shim_uname.c
```
m1ghtym0, line 89:
```suggestion
	// Taken from: https://github.com/gramineproject/gramine/blob/c83ec08f10cdbb3a258d18b118dd95602a55abc9/libos/src/sys/libos_uname.c
```
Seems like Gramine also updated their uname implementation. nodename is now a PAL-provided value.

m1ghtym0 (changes requested):
Thanks for noticing! Please also update the reference to the gramine implementation.

m1ghtym0 (approved):

PR #336, opened 2022-12-15
Add delete operations to StdStore

daniel-weisse:
### Proposed changes
- Add delete operation to StdStore to delete entries from store
  - `Delete()` is not part of the regular `store.Store` interface
  - I decided to add `Delete` to `store.Transaction` to save on some type assertions when using store transaction

### DIFF
--- a/coordinator/store/stdstore/stdstore.go
+++ b/coordinator/store/stdstore/stdstore.go
@@ -49,12 +49,25 @@ func (s *StdStore) Get(request string) ([]byte, error) {
 
 // Put saves a value in StdStore by Type and Name.
 func (s *StdStore) Put(request string, requestData []byte) error {
-	tx, err := s.BeginTransaction()
+	tx, err := s.beginTransaction()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


coordinator/store/stdstore/stdstore.go
```diff
@@ -122,6 +126,19 @@ func (s *StdStore) SetEncryptionKey(encryptionKey []byte) error {
 	return s.sealer.SetEncryptionKey(encryptionKey)
 }
 
+func (s *StdStore) beginTransaction() (*StdTransaction, error) {
```
m1ghtym0, line 129: Moving that into a private function for style reasons?

m1ghtym0 (approved):

coordinator/store/stdstore/stdstore.go
daniel-weisse: Moved to private so we can work on `*StdTransaction` directly instead of the `store.Transaction` interface

PR #339, opened 2022-12-21
Create integration testing framework

daniel-weisse:
### Proposed changes
- Refactor the integration test to use a framework like approach
  - This allows reuse of the old integration test functions like starting the Coordinator, setting a manifest etc., without having to directly add new functions to `test/integration_test.go`

### DIFF
--- a/test/framework/framework.go
+++ b/test/framework/framework.go
@@ -401,24 +401,7 @@ func (i IntegrationTest) StartMarbleClient(cfg MarbleConfig) bool {
 }
 
 // TriggerRecovery starts a Coordinator, sets the manifest, and triggers a recovery.
-func (i IntegrationTest) TriggerRecovery(manifest manifest.Manifest) ([]byte, *os.Process, *os.Process, CoordinatorConfig, MarbleConfig, string) {
-	// start Coordinator
-	log.Println("Starting a coordinator enclave")
-	cfg := NewCoordinatorConfig()
-	coordinatorProc := i.StartCoordinator(cfg)
-	i.require.NotNil(coordinatorProc)
-
-	// set Manifest
-	log.Println("Setting the Manifest")
-	recoveryResponse, err := i.SetManifest(manifest)
-	i.require.NoError(err, "failed to set Manifest")
-
-	// start server
-	log.Println("Starting a Server-Marble")
-	serverCfg := NewMarbleConfig(i.MeshServerAddr, "testMarbleServer", "server,backend,localhost")
-	serverProc := i.StartMarbleServer(serverCfg)
-	i.require.NotNil(serverProc, "failed to start server-marble")
-
+func (i IntegrationTest) TriggerRecovery(coordinatorCfg CoordinatorConfig, coordinatorProc *os.Process) (*os.Process, string) {
 	// get certificate
 	log.Println("Save certificate before we try to recover.")
 	client := http.Client{Transport: i.transportSkipVerify}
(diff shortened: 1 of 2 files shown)


test/framework/framework.go
```diff
@@ -0,0 +1,537 @@
...
+
+	panic(err.Error() + "\n" + string(out))
+}
+
+// TriggerRecovery starts a Coordinator, sets the manifest, and triggers a recovery.
+func (i IntegrationTest) TriggerRecovery(manifest manifest.Manifest) ([]byte, *os.Process, *os.Process, CoordinatorConfig, MarbleConfig, string) {
+	// start Coordinator
+	log.Println("Starting a coordinator enclave")
+	cfg := NewCoordinatorConfig()
+	coordinatorProc := i.StartCoordinator(cfg)
```
m1ghtym0, line 408: Why is this starting a new Coordiantor, and all other functions assume it's already running?

daniel-weisse: (Please note that this PR only changes the existing code to run as methods of `IntegrationTest`)

I guess this is mostly comes down to implementation decisions when the recover test was created.
Imo the functions `TestRecoveryRestoreKey`, and `TestRecoveryReset` could be changed to run some of the tasks from the `TriggerRecovery` method

m1ghtym0: Do you plan to do further refactoring in separate PRs?

daniel-weisse: This PR was intended to provide us with an easier way to add new integration tests and adapt existing ones in the future.
Might be a good idea to refactor the recover tests in this PR

m1ghtym0 (approved):

test/framework/framework.go
daniel-weisse: Refactored the code so the test functions are responsible for starting the coordinator

m1ghtym0 (approved):
Thanks, looks better now:-)

test/framework/framework.go
```diff
@@ -0,0 +1,520 @@
...
+	}
+
+	if _, ok := err.(*exec.ExitError); ok {
+		return false
+	}
+
+	panic(err.Error() + "\n" + string(out))
+}
+
+// TriggerRecovery starts a Coordinator, sets the manifest, and triggers a recovery.
```
thomasten, line 403:
```suggestion
// TriggerRecovery triggers a recovery.
```

test/integration_test.go
```diff
@@ -87,144 +67,126 @@ func TestMain(m *testing.M) {
...
-
-	// Adjust unit test update manifest to work with the integration test
-	updatedManifest.Packages["backend"] = updatedManifest.Packages["frontend"]
-}
-
 // sanity test of the integration test environment
 func TestTest(t *testing.T) {
 	assert := assert.New(t)
+	f := framework.New(t, *buildDir, simFlag, *noenclave, marbleTestAddr, meshServerAddr, clientServerAddr, IntegrationManifestJSON, UpdateManifest)
+	f.UpdateManifest()
```
thomasten, line 74: Maybe replace all occurrences of this with
```suggestion
	f := newFramework(t)
```
with newFramework like
```go
func newFramework(t *testing.T) *framework.IntegrationTest {
	f := framework.New(t, *buildDir, simFlag, *noenclave, marbleTestAddr, meshServerAddr, clientServerAddr, IntegrationManifestJSON, UpdateManifest)
	f.UpdateManifest()
	return f
}
```
?
Or keep as is if you don't like it.

thomasten (approved):

PR #342, opened 2023-01-02
docs: multi-party manifest update

thomasten:
add documentation for multi-party manifest update to the public docs


m1ghtym0 (commented):
The changes look good, but I'm missing a few things from a user perspective (that might not be implemented yet):
* How can I check if an update procedure is happening and whose ack is missing (e.g. `marblerun status`)
* How can I cancel an update procedure
* Explain that the acknowledge process means uploading the "same" manifest

PR #344, opened 2023-01-04
Allow TCB levels to be accepted besides UpToDate

OverOrion:
### Proposed changes
This PR aims to give users the ability to accept different TCB levels. This can be useful in cases where the grace period has ended but for some reason (e.g. the BIOS update has not been relased yet) SWHardening could not take place.

### Additional info
Currently a user needs to set **both** the command line flags and in the manifest as well. The command line flag was necessary for me as I could not find a way to extract the manifest values and inject them to the call site without major refactoring. Could you please give me some hints regarding this?

I was unable to run E2E tests via Kubernetes deployment because it seemed to me that some package (ego? era? not sure) were using the upstream version instead of my patched branch (I was using the provided Dockerfile in redis-gramine). This is likely a fault on my side, but wanted to mention it nevertheless. I would appreciate it if you could run E2E on it before merging.

### DIFF
--- a/cli/cmd/certificateChain.go
+++ b/cli/cmd/certificateChain.go
@@ -24,7 +24,7 @@ func newCertificateChain() *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			hostName := args[0]
-			return cliCertificateChain(hostName, certFilename, eraConfig, insecureEra)
+			return cliCertificateChain(hostName, certFilename, eraConfig, insecureEra, acceptedTCBStatuses)
 		},
 		SilenceUsage: true,
 	}
@@ -35,8 +35,8 @@ func newCertificateChain() *cobra.Command {
 }
 
 // cliCertificateChain gets the certificate chain of the MarbleRun Coordinator.
-func cliCertificateChain(host string, output string, configFilename string, insecure bool) error {
-	certs, err := verifyCoordinator(host, configFilename, insecure)
+func cliCertificateChain(host string, output string, configFilename string, insecure bool, acceptedTCBStatuses []string) error {
+	certs, err := verifyCoordinator(host, configFilename, insecure, acceptedTCBStatuses)
 	if err != nil {
 		return err
 	}
--- a/cli/cmd/certificateIntermediate.go
+++ b/cli/cmd/certificateIntermediate.go
@@ -24,7 +24,7 @@ func newCertificateIntermediate() *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			hostName := args[0]
-			return cliCertificateIntermediate(hostName, certFilename, eraConfig, insecureEra)
+			return cliCertificateIntermediate(hostName, certFilename, eraConfig, insecureEra, acceptedTCBStatuses)
 		},
 		SilenceUsage: true,
 	}
@@ -35,8 +35,8 @@ func newCertificateIntermediate() *cobra.Command {
 }
 
 // cliCertificateIntermediate gets the intermediate certificate of the MarbleRun Coordinator.
-func cliCertificateIntermediate(host string, output string, configFilename string, insecure bool) error {
-	certs, err := verifyCoordinator(host, configFilename, insecure)
+func cliCertificateIntermediate(host string, output string, configFilename string, insecure bool, acceptedTCBStatuses []string) error {
+	certs, err := verifyCoordinator(host, configFilename, insecure, acceptedTCBStatuses)
 	if err != nil {
 		return err
 	}
--- a/cli/cmd/certificateRoot.go
+++ b/cli/cmd/certificateRoot.go
@@ -24,7 +24,7 @@ func newCertificateRoot() *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			hostName := args[0]
-			return cliCertificateRoot(hostName, certFilename, eraConfig, insecureEra)
+			return cliCertificateRoot(hostName, certFilename, eraConfig, insecureEra, acceptedTCBStatuses)
 		},
 		SilenceUsage: true,
 	}
@@ -35,9 +35,9 @@ func newCertificateRoot() *cobra.Command {
 }
 
 // cliCertificateRoot gets the root certificate of the MarbleRun Coordinator and saves it to a file.
-func cliCertificateRoot(host string, output string, configFilename string, insecure bool) error {
+func cliCertificateRoot(host string, output string, configFilename string, insecure bool, acceptedTCBStatuses []string) error {
 	var certs []*pem.Block
-	certs, err := verifyCoordinator(host, configFilename, insecure)
+	certs, err := verifyCoordinator(host, configFilename, insecure, acceptedTCBStatuses)
 	if err != nil {
 		return err
 	}
--- a/cli/cmd/manifestGet.go
+++ b/cli/cmd/manifestGet.go
@@ -31,7 +31,7 @@ Optionally get the manifests signature or merge updates into the displayed manif
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			hostName := args[0]
-			cert, err := verifyCoordinator(hostName, eraConfig, insecureEra)
+			cert, err := verifyCoordinator(hostName, eraConfig, insecureEra, acceptedTCBStatuses)
 			if err != nil {
 				return err
 			}
--- a/cli/cmd/util.go
+++ b/cli/cmd/util.go
@@ -20,7 +20,9 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/edgelesssys/ego/attestation"
 	"github.com/edgelesssys/era/era"
+	"github.com/edgelesssys/era/util"
 	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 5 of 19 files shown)


m1ghtym0:
Thank you @OverOrion!
This is a feature we had on the roadmap, perfect timing with your PR:-)

cli/cmd/certificateChain.go
```diff
@@ -24,19 +24,20 @@ func newCertificateChain() *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			hostName := args[0]
-			return cliCertificateChain(hostName, certFilename, eraConfig, insecureEra)
+			return cliCertificateChain(hostName, certFilename, eraConfig, insecureEra, acceptedTcbLevels)
 		},
 		SilenceUsage: true,
 	}
 
 	cmd.Flags().StringVarP(&certFilename, "output", "o", "marblerunChainCA.crt", "File to save the certificate to")
+	cmd.PersistentFlags().StringSliceVar(&acceptedTcbLevels, "accepted-tcb-levels", []string{"UpToDate"}, "Coma seperated list of user accepted TCB levels (e.g. ConfigurationNeeded,ConfigurationAndSWHardeningNeeded)")
```
daniel-weisse, line 33:
```suggestion
```
Don't need to define the flag again since it is already defined by the parent command.

cli/cmd/certificateIntermediate.go
```diff
@@ -24,19 +24,20 @@ func newCertificateIntermediate() *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			hostName := args[0]
-			return cliCertificateIntermediate(hostName, certFilename, eraConfig, insecureEra)
+			return cliCertificateIntermediate(hostName, certFilename, eraConfig, insecureEra, acceptedTcbLevels)
 		},
 		SilenceUsage: true,
 	}
 
 	cmd.Flags().StringVarP(&certFilename, "output", "o", "marblerunIntermediateCA.crt", "File to save the certificate to")
+	cmd.PersistentFlags().StringSliceVar(&acceptedTcbLevels, "accepted-tcb-levels", []string{"UpToDate"}, "Coma seperated list of user accepted TCB levels (e.g. ConfigurationNeeded,ConfigurationAndSWHardeningNeeded)")
```
daniel-weisse, line 33:
```suggestion
```
Same here

cli/cmd/certificateRoot.go
```diff
@@ -24,20 +24,21 @@ func newCertificateRoot() *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			hostName := args[0]
-			return cliCertificateRoot(hostName, certFilename, eraConfig, insecureEra)
+			return cliCertificateRoot(hostName, certFilename, eraConfig, insecureEra, acceptedTcbLevels)
 		},
 		SilenceUsage: true,
 	}
 
 	cmd.Flags().StringVarP(&certFilename, "output", "o", "marblerunRootCA.crt", "File to save the certificate to")
+	cmd.PersistentFlags().StringSliceVar(&acceptedTcbLevels, "accepted-tcb-levels", []string{"UpToDate"}, "Coma seperated list of user accepted TCB levels (e.g. ConfigurationNeeded,ConfigurationAndSWHardeningNeeded)")
```
daniel-weisse, line 33:
```suggestion
```
And here

cli/cmd/manifestGet.go
```diff
@@ -61,6 +61,7 @@ Optionally get the manifests signature or merge updates into the displayed manif
 	cmd.Flags().BoolVarP(&signature, "signature", "s", false, "Set to additionally display the manifests signature")
 	cmd.Flags().BoolVarP(&displayUpdate, "display-update", "u", false, "Set to merge updates into the displayed manifest")
 	cmd.Flags().StringVarP(&output, "output", "o", "", "Save output to file instead of printing to stdout")
+	cmd.PersistentFlags().StringSliceVar(&acceptedTcbLevels, "accepted-tcb-levels", []string{"UpToDate"}, "Coma seperated list of user accepted TCB levels (e.g. ConfigurationNeeded,ConfigurationAndSWHardeningNeeded)")
```
daniel-weisse, line 64:
```suggestion
```
And here also

cli/cmd/util.go
```diff
@@ -58,7 +61,11 @@ func verifyCoordinator(host string, configFilename string, insecure bool) ([]*pe
 
 	// get certificate using provided config
 	if configFilename != "" {
-		pemBlock, _, err := era.GetCertificate(host, configFilename)
+		pemBlock, tcbLevel, err := era.GetCertificate(host, configFilename)
+		if err == attestation.ErrTCBLevelInvalid && util.StringSliceContains(acceptedTcbLevels, tcbLevel.String()) {
```
daniel-weisse, line 65:
```suggestion
		if errors.Is(err, attestation.ErrTCBLevelInvalid) && util.StringSliceContains(acceptedTcbLevels, tcbLevel.String()) {
```
A bit cleaner, since this will also return true if the error contains a wrapped `ErrTCBLevelInvalid` error

coordinator/quote/ert.go
```diff
@@ -26,6 +26,8 @@ type PackageProperties struct {
 	ProductID *uint64
 	// Security version number of the package.
 	SecurityVersion *uint
+	// Accepted TCB levels
+	AcceptedTcbLevels []string
```
daniel-weisse, line 30:
```suggestion
	AcceptedTCBLevels []string
```
Just a small naming thing

coordinator/quote/ertvalidator/ertvalidator.go
```diff
@@ -30,7 +32,11 @@ func (m *ERTValidator) Validate(givenQuote []byte, cert []byte, pp quote.Package
 	// Verify Quote
 	report, err := enclave.VerifyRemoteReport(givenQuote)
 	if err != nil {
-		return fmt.Errorf("verifying quote failed: %v", err)
+		if err == attestation.ErrTCBLevelInvalid && util.StringSliceContains(pp.AcceptedTcbLevels, report.TCBStatus.String()) {
```
daniel-weisse, line 35:
```suggestion
		if errors.Is(err, attestation.ErrTCBLevelInvalid) && util.StringSliceContains(pp.AcceptedTcbLevels, report.TCBStatus.String()) {
```

daniel-weisse (changes requested):
Changes look good to me, thanks
Just noted down some small style changes

Regarding having to set both CLI flags and manifest:
This is OK, and imo the correct way to implement this.

daniel-weisse:
>I was unable to run E2E tests via Kubernetes deployment because it seemed to me that some package (ego? era? not sure) were using the upstream version instead of my patched branch (I was using the provided Dockerfile in redis-gramine).

The dockerfile builds the premain from latest upstream source.

If you want to use a premain build from your changes instead, you can adjust the Dockerfile to pull from your repo
```diff
4c4
< RUN git clone https://github.com/edgelesssys/marblerun.git /marblerun
---
> RUN git clone --branch tcb-allow-list-pr https://github.com/OverOrion/marblerun.git /marblerun
```

Alternatively, you could also just copy the built premain into the container directly, instead of building it in docker:
```diff
37c37
< COPY --from=build-premain /premain/build/premain-libos /gramine/CI-Examples/redis/
---
> COPY ./premain-libos /gramine/CI-Examples/redis/
```

OverOrion:
> Changes look good to me, thanks Just noted down some small style changes
> 
> Regarding having to set both CLI flags and manifest: This is OK, and imo the correct way to implement this.

Thank you for the quick review, I have addressed them and did a force push (it seemed like the way how it is expected to be done in previous PRs).

coordinator/quote/ertvalidator/ertvalidator.go
```diff
@@ -13,7 +13,9 @@ import (
 	"encoding/hex"
 	"fmt"
```
daniel-weisse, line 14:
```suggestion
	"encoding/hex"
	"errors"
	"fmt"
```
missing import of `errors` package

OverOrion: Thank you, sorry for that, fixed in the last force push.

coordinator/quote/ert.go
```diff
@@ -26,6 +26,8 @@ type PackageProperties struct {
 	ProductID *uint64
 	// Security version number of the package.
 	SecurityVersion *uint
+	// Accepted TCB levels
+	AcceptedTCBLevels []string
```
thomasten, line 30: Maybe we should call this AcceptedTCBStatuses because it's not only about the level, but also about configuration and sw hardening.

thomasten, line 30: I wonder if we should define this per package or globally in the manifest. I don't have a strong opinion. What do you think? @OverOrion, @daniel-weisse, @m1ghtym0

daniel-weisse: I think having the option to define this per package makes sense, since applications might be running on different systems (but are verified by the same instance of MarbleRun). There could be applications where a certain TCB level is acceptable, which might not be the case for other applications.

OverOrion: This is what I had in mind as well.

OverOrion: You are right, I will rename it, thanks for the note. Should I do it in a new separate commit or squash it?

thomasten: If it's ok for you to edit both of your commits (it should be renamed in the CLI, too), please do so. If that's too much effort, you can also do a separate commit.

m1ghtym0: Yes, I agree.

thomasten (approved):

daniel-weisse (approved):

PR #346, opened 2023-01-09
Update samples and PreMain for Gramine v1.3 compatibility

daniel-weisse:
### Proposed changes
Went over our samples and tried to make them functional for v1.3.
Ran into a couple issues with the nginx sample, but they are fixed now.

I have also updated the Gramine PreMain to now be compatible with the new encrypted key format, which was first introduced in [Gramine v1.2](https://github.com/gramineproject/gramine/releases/tag/v1.2), and squashed a "bug" where the protected files key was written twice.

### Additional info
I marked the gramine-prepare command as deprecated for now, since usage seems to be low/non existent, and keeping it up to date or compatible with the latest Gramine manifest syntax seems to be more of a hassle than its worth.


docs/docs/building-services/gramine.md
```diff
@@ -37,32 +47,46 @@ loader.env.EDG_MARBLE_DNS_NAMES = { passthrough = true }
...
 ### Secret files
+
 A Marble's secrets, e.g. a certificate and private key, can be provisioned as files. You can utilize Gramine's in-memory filesystem [`tmpfs`](https://gramine.readthedocs.io/en/latest/manifest-syntax.html#fs-mount-points), so the secrets will never show up on the host's file system :
+
 ```toml
-fs.mount.secrets.type = "tmpfs"
-fs.mount.secrets.path = "/secrets"
+fs.mounts = [
+    # ...
+    { type = "tmpf", path = "/secrets" },
```
Nirusu, line 83:
```suggestion
    { type = "tmpfs", path = "/secrets" },
```
I suppose that's a typo, right?

samples/gramine-hello/README.md
```diff
@@ -42,5 +42,5 @@ curl -k --data-binary @manifest.json https://localhost:4433/manifest
 The type of the Marble is defined in the `manifest.json`. In this example, the manifest defines a single Marble, which is called "hello". To run the application, you need to set the `EDG_MARBLE_TYPE` environement variable to that name.
 
 ```sh
-EDG_MARBLE_TYPE=hello make run
+make run
```
Nirusu, line 45: I think the reason this was external before was to highlight that you could easily change the Marble type here to your own setup? 

This would make it slightly hidden this way and would be fine if you follow the steps 1:1, but if you use it as a template to port your own application, it could be slightly confusing.

samples/gramine-hello/README.md
```diff
@@ -42,5 +42,5 @@ curl -k --data-binary @manifest.json https://localhost:4433/manifest
 The type of the Marble is defined in the `manifest.json`. In this example, the manifest defines a single Marble, which is called "hello". To run the application, you need to set the `EDG_MARBLE_TYPE` environement variable to that name.
```
Nirusu, line 42: Can you actually override an environment variable defined inside the `Makefile` from the outside by e.g.:
`EDG_MARBLE_TYPE=myownmarble make run`
?

If no, you may want to adjust this sentence to highlight that the definition is in the `Makefile`.

docs/docs/building-services/gramine.md
```diff
@@ -74,21 +98,32 @@ You can specify the files' content in the MarbleRun manifest:
 ...
 ```
 
-Note that Gramine also allows to store files encrypted on the host's file system. The so called `protected files` require to initialize the protected files key by writing it hex-encoded to the virtual `protected_files_key` device:
+Note that Gramine also allows to store files encrypted on the host's file system. The so encrypted filesystem requires to initialize an encrypted files key by writing it to the virtual `protected_files_key` device.
+For debugging purposes, you can define the key directly in the Gramine manifest.
 
+```toml
+fs.insecure__keys.[KEY_NAME]] = "[16-byte hex value]"
```
Nirusu, line 105:
```suggestion
fs.insecure__keys.[KEY_NAME]] = "[32-character hex value]"
```
I know this existed before, but I would just call it the same way as Gramine does (32 character hex value).

Of course 32 chars hex string == 16-byte hex value but the latter can be slightly ambiguous if you switch between the two docs.

docs/docs/building-services/gramine.md
```diff
@@ -74,21 +98,32 @@ You can specify the files' content in the MarbleRun manifest:
...
+  { type = "encrypted", path = "/secrets", uri = "file:/path/to/local/directory", key_name = "[KEY_NAME] },
+  # ...
+]
 ```
-sgx.protected_files_key = "[16-byte hex value]"
-sgx.protected_files.[identifier] = "[URI]"
-```
-You can see how this key can be initialized with MarbleRun in the [nginx example](https://github.com/edgelesssys/marblerun/tree/master/samples/gramine-nginx).
+
+MarbleRun allows provisioning the encrypted files key at runtime to your enclave.
```
Nirusu, line 113: I would maybe highlight it a bit more that this is actually the recommended way by Gramine and `insecure__keys` is just here for a quick-start and minimal sample.

Nirusu (commented):
Thanks for updating all the samples. Generally LGTM, just a typo (?) I found and a few minor things I would maybe check or change, but nothing really important.

samples/gramine-hello/README.md
daniel-weisse: See: https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/947

samples/gramine-hello/README.md
daniel-weisse: Maybe we should just remove that sentence. Does not make a lot of sense if we already set default values

samples/gramine-hello/README.md
daniel-weisse: Or we remove the `make run` and replace it with `EDG_MARBLE_TYPE=hello gramine-sgx hello`.
This way the step is consistent with our other gramine samples

daniel-weisse: kept the make target for now, but adjusted the README to only use `EDG_MARBLE_TYPE=hello gramine-sgx hello`
:+1:

docs/docs/building-services/gramine.md
```diff
@@ -74,21 +98,32 @@ You can specify the files' content in the MarbleRun manifest:
 ...
 ```
 
-Note that Gramine also allows to store files encrypted on the host's file system. The so called `protected files` require to initialize the protected files key by writing it hex-encoded to the virtual `protected_files_key` device:
+Note that Gramine also allows to store files encrypted on the host's file system. The so encrypted filesystem requires to initialize an encrypted files key by writing it to the virtual `protected_files_key` device.
```
Nirusu, line 101:
```suggestion
Note that Gramine also allows to store files encrypted on the host's file system. This is done by providing one or multiple encryption key and defining a mount point as encrypted in combination with the key to be used.
```

Or something simpler, but `protected_files_key` is deprecated and does not need to be mentioned here for the "for debug purpose" case.

docs/docs/building-services/gramine.md
```diff
@@ -74,21 +98,32 @@ You can specify the files' content in the MarbleRun manifest:
...
+  { type = "encrypted", path = "/secrets", uri = "file:/path/to/local/directory", key_name = "[KEY_NAME] },
+  # ...
+]
 ```
-sgx.protected_files_key = "[16-byte hex value]"
-sgx.protected_files.[identifier] = "[URI]"
-```
-You can see how this key can be initialized with MarbleRun in the [nginx example](https://github.com/edgelesssys/marblerun/tree/master/samples/gramine-nginx).
+
+For production use cases, MarbleRun can provision your enclave with encrypted file keys at runtime.
```
Nirusu, line 113:
```suggestion
For production use cases, MarbleRun can provision your enclave with encrypted file keys at runtime by creating a virtual file under `/dev/attestation/keys/<key_name>` which Graphene can ingest for key provisioning.
```

samples/gramine-hello/README.md
Nirusu: Seems like a good compromise and also reduces the use of a magic Makefile. I like it 👍🏻

docs/docs/building-services/gramine.md
daniel-weisse: Shortened the section by referring to the Gramine docs where possible

Nirusu: Generally that's fine, though I would still keep the distinction between the "debug" use and the proper key deployment with MarbleRun. The `insecure` should already give a hint, but since it's important for security better be safe than sorry I guess. 

Also, one might wonder why we even advertise the insecure way. Keeping the distinction between "quick for debug" versus recommendation for production use cases also allows us to show the value MarbleRun provides here (seamless key provisioning).

But that's the last nit I have, don't worry :)

daniel-weisse: I would prefer to keep our docs on Gramine as minimal as possible, since Gramine's manifest syntax can, and does, change from one release to another.
Users should not need our docs to learn about, or understand Gramine's features.
The fact that encryption keys should be provisioned at runtime by an external application is already mentioned in their docs, and I think they explain it quite well.

Nirusu: Oh never mind, I actually didn't fully realize in the diff that you removed the insecure example from the docs.
Sorry, too early in the morning. That's fine then :)

Nirusu (approved):

docs/docs/building-services/gramine.md
```diff
@@ -1,34 +1,46 @@
...
 ]
 ```
+
 After the premain is done running, it will automatically spawn your application.
 
 ### Host environment variables
-The premain needs access to some host [environment variables for configuration](../workflows/add-service.md#step-3-start-your-service):
+
+By default, environment variables from the host won't be passed to the application.
+Gramine allows to [pass through select environment variables from the host](https://gramine.readthedocs.io/en/v1.3/manifest-syntax.html#environment-variables).
```
m1ghtym0, line 41:
```suggestion
Gramine allows to [pass through whitelisted environment variables from the host](https://gramine.readthedocs.io/en/v1.3/manifest-syntax.html#environment-variables).
```

m1ghtym0 (approved):

PR #348, opened 2023-01-13
Fix default configuration and acceptedTCBStatuses

OverOrion:
### Proposed changes
This PR fixes the case when a user does not supply the eraconfig explicitly but has the flag `accepted-tcb-statuses` set. This flag only worked if it was explicitly supplied.

### DIFF
--- a/cli/cmd/util.go
+++ b/cli/cmd/util.go
@@ -110,7 +110,11 @@ func verifyCoordinator(host string, configFilename string, insecure bool, accept
 		return nil, err
 	}
 
-	pemBlock, _, err := era.GetCertificate(host, eraDefaultConfig)
+	pemBlock, tcbStatus, err := era.GetCertificate(host, eraDefaultConfig)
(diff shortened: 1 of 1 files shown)


daniel-weisse (approved):
Thanks, looks good to me

cli/cmd/util.go
```diff
@@ -51,58 +51,67 @@ var (
...
+		configFile = configFilename
+	} else {
+		// get latest config from github if none specified
+		if err := fetchLatestCoordinatorConfiguration(); err != nil {
+			return nil, err
+		}
+		configFile = eraDefaultConfig
+	}
+
+	pemBlock, tcbStatus, err := era.GetCertificate(host, configFile)
```
thomasten, line 110:
```suggestion
	if configFilename == "" {
		// get latest config from github if none specified
		if err := fetchLatestCoordinatorConfiguration(); err != nil {
			return nil, err
		}
		configFilename = eraDefaultConfig
	}

	pemBlock, tcbStatus, err := era.GetCertificate(host, configFilename)
```
nit: IMO this is more readable without introducing a new variable. But if you disagree, we could also keep it as is.

OverOrion: Thanks, you are right! :100:

PR #349, opened 2023-01-13
Fix example gramine redis Dockerfile

OverOrion:
### Proposed changes
The `COPY ./premain-libos /premain/build/premain-libos` is not required as this artifact will be built by the previous command.
Also removed a duplicate package in the `apt-get install` step.


daniel-weisse (approved):
Thanks, seem to have slipped through with my last PR :sweat_smile:

PR #354, opened 2023-01-20
Access token flag

daniel-weisse:
### Proposed changes
- Add the `enterprise-access-token` flag to configure a pull secret and deploy an enterprise Coordinator
- Reduce liveness and readiness probe frequency from once every 10 seconds to 60 seconds


m1ghtym0 (commented):
Conceptionally LGTM:-)
I don't have a strong opinion on the naming/flag format.

m1ghtym0 (approved):

PR #362, opened 2023-01-30
add reproducible build and production build

thomasten:


dockerfiles/start.sh
```diff
@@ -2,13 +2,9 @@
 
 if [[ "${DCAP_LIBRARY}" == "intel" ]]
 then
-    # rename the library installed by az-dcap-client
-    mv /usr/lib/libdcap_quoteprov.so /usr/lib/libdcap_quoteprov.so.azure
-    # create a link to the intel quote provider library
-    ln -s /usr/lib/x86_64-linux-gnu/dcap/libdcap_quoteprov.so.intel /usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so
-    ln -s /usr/lib/x86_64-linux-gnu/dcap/libdcap_quoteprov.so.intel /usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so.1
+    apt-get install -qq libsgx-dcap-default-qpl
```
daniel-weisse, line 5:
```suggestion
    apt-get install -qq libsgx-dcap-default-qpl > /dev/null
```

There is still *some* output even with `-qq`

build_with_version.cmake
```diff
@@ -5,7 +5,6 @@ execute_process(COMMAND bash "-c" "git rev-parse HEAD | tr -d '\n'"
 set(COMPILER ${CMAKE_ARGV3})
 set(PROJECT_VERSION ${CMAKE_ARGV4})
 set(OUTPUT_NAME ${CMAKE_ARGV5})
-set(BUILD_SOURCE ${CMAKE_ARGV6})
 set(INJECT_PATH ${CMAKE_ARGV7})
 set(TRIMPATH ${CMAKE_ARGV8})
```
daniel-weisse, line 9:
```suggestion
set(INJECT_PATH ${CMAKE_ARGV6})
set(TRIMPATH ${CMAKE_ARGV7})
```
These probably need to be adjusted if we remove argv[6]

dockerfiles/start.sh
thomasten: Do you think the output is undesirable? I don't have an opinion on this. Will follow your suggestion if you confirm.

build_with_version.cmake
thomasten: good catch!

dockerfiles/start.sh
daniel-weisse: I think its a bit unsightly to get the "Reading package database" messages every time you start the container.
We don't lose anything by hiding it, but logs look a bit messy if we do.

dockerfiles/start.sh
```diff
@@ -2,7 +2,7 @@
 
 if [[ "${DCAP_LIBRARY}" == "intel" ]]
 then
-    apt-get install -qq libsgx-dcap-default-qpl
+    apt-get install -qq libsgx-dcap-default-qpl > /dev/null
 else
     apt-get install -qq az-dcap-client
```
daniel-weisse, line 7:
```suggestion
    apt-get install -qq az-dcap-client > /dev/null
```
We should silence output for both commands ideally

dockerfiles/start.sh
```diff
@@ -2,13 +2,9 @@
 
 if [[ "${DCAP_LIBRARY}" == "intel" ]]
 then
-    # rename the library installed by az-dcap-client
-    mv /usr/lib/libdcap_quoteprov.so /usr/lib/libdcap_quoteprov.so.azure
-    # create a link to the intel quote provider library
-    ln -s /usr/lib/x86_64-linux-gnu/dcap/libdcap_quoteprov.so.intel /usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so
-    ln -s /usr/lib/x86_64-linux-gnu/dcap/libdcap_quoteprov.so.intel /usr/lib/x86_64-linux-gnu/libdcap_quoteprov.so.1
+    apt-get install -qq libsgx-dcap-default-qpl > /dev/null
```
daniel-weisse, line 5:
```suggestion
    apt-get install -qq libsgx-dcap-default-qpl > /dev/null 2>&1
```
Did some quick testing, if we want to silence all output we need to redirect stdout and stderr to /dev/null

dockerfiles/start.sh
```diff
@@ -2,13 +2,9 @@
...
 else
-    export AZDCAP_DEBUG_LOG_LEVEL="${AZDCAP_DEBUG_LOG_LEVEL:=ERROR}"
+    apt-get install -qq az-dcap-client > /dev/null
```
daniel-weisse, line 7:
```suggestion
    apt-get install -qq az-dcap-client > /dev/null 2>&1
```

daniel-weisse (approved):

PR #364, opened 2023-01-31
Update image name in workflow

daniel-weisse:
### Proposed changes
- Container image names were updated with v1.0.0 from `ghcr.io/edgelesssys/<app>` to `ghcr.io/edgelesssys/marblerun/<app>`, this PR fixes the workflow to push images using the new naming scheme

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -65,7 +65,7 @@ jobs:
           -d '{"event_type": "docker-build",
               "client_payload":{"repository":"marblerun",
                                 "sign":"nightly",
-                                "imagename":"coordinator",
+                                "imagename":"marblerun/coordinator",
                                 "tag":"nightly",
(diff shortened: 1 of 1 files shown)


.github/workflows/unittests.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           -d '{"event_type": "docker-build",
               "client_payload":{"repository":"marblerun",
                                 "sign":"nightly",
-                                "imagename":"coordinator",
+                                "imagename":"marblerun/coordinator",
```
thomasten, line 68:
```suggestion
                                "imagename":"marblerun/coordinator-debug",
```
I think it will build a debug coordinator by default? Also seems reasonable to me to publish nightlies (only) for the debug image.

daniel-weisse: Sounds good

PR #367, opened 2023-02-08
Charts: fix missing provision in resources.limits

OverOrion:
Without this patch it is not possible to deploy marblerun using simple helm commands, it would fail with:
```
[error_driver2api sgx_enclave_common.cpp:261] Enclave not authorized to run, .e.g. provisioning enclave hosted in app without access rights to /dev/sgx_provision. You need add the user id to │
│ [load_pce ../pce_wrapper.cpp:188] Error, call sgx_create_enclave for PCE fail [load_pce], SGXError:4004.  
```


OverOrion:
> As already mentioned can you add `util.IntelProvision.String()` in the [install command](https://github.com/edgelesssys/marblerun/blob/7eb8b4d5ecf433a3fa960232a956bee42260a0cd/cli/cmd/install.go#L428) to make sure the provision key is removed if a user decides to use a device plugin that does not use the `sgx.intel.com` resources keys?
> 
> Other wise looks good to me, thanks +1

Yes, thank you for your guidance, squashed it into the previous commit.

daniel-weisse (approved):

PR #368, opened 2023-02-09
Enable MarbleRun Kubernetes installation using only helm

daniel-weisse:
### Proposed changes
- Add cert-manager self signed certificate for marble-injector webhook
  - Allows standalone helm installation, no MarbleRun CLI needed
  - Requires a change in marble-injector cmd line flags. Installing this chart with and older CLI will cause some errors  if the chart is not manually edited.
- Add release namespace name to mutating webhook configuration name of marble-injector
  - Allows installing multiple MarbleRun charts into the same cluster without causing webhook configuration collisions
- Add `objectSelector` and `namespaceSelector` to values file
  - Allows easily updating the mutation triggers of the webhook

PR #371, opened 2023-02-13
Allow multiple marblerun instances to be installed with helm

OverOrion:
### Proposed changes
Based on #368 this PR allows multiple Marblerun deployments in the same Kubernetes cluster, but with different namespaces.

~Currently the user needs to create the `namespace` and create the secret manually:~
EDIT: it works without manually creating it.

### How the `charts/README.md` should be edited?
We used the following helm command to install marblerun:
```bash
helm install marblerun ./charts/ --create-namespace -n $namespace \
        --set marbleInjector.useCertManager=true \
        --set marbleInjector.start=true \
        --set coordinator.hostname=$host
        --set dcap.qpl=intel \
        --set dcap.pccsUrl=https://$host:8081/sgx/certification/v3/ \
        --set dcap.useSecureCert=FALSE \
        -f ./charts/values.yaml 

```
Should it be added to the `README.md`?

### DIFF
--- a/charts/templates/coordinator.yaml
+++ b/charts/templates/coordinator.yaml
@@ -23,7 +23,7 @@ spec:
         app.kubernetes.io/component: coordinator
         app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
         app.kubernetes.io/name: coordinator
-        app.kubernetes.io/part-of: {{ .Release.Namespace }}
+        app.kubernetes.io/part-of: marblerun
         app.kubernetes.io/version: {{ .Values.coordinator.version }}
         {{ .Values.global.coordinatorComponentLabel }}: coordinator
         {{ .Values.global.coordinatorNamespaceLabel }}: {{ .Release.Namespace }}
@@ -110,7 +110,7 @@ metadata:
     app.kubernetes.io/component: persistent-storage
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator-pv-claim
-    app.kubernetes.io/part-of: {{ .Release.Namespace }}
+    app.kubernetes.io/part-of: marblerun
     app.kubernetes.io/version: {{ .Values.coordinator.version }}
 spec:
   {{- with .Values.coordinator.storageClass }}
@@ -131,7 +131,7 @@ metadata:
     app.kubernetes.io/component: client-api
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator-client-api
-    app.kubernetes.io/part-of: {{ .Release.Namespace }}
+    app.kubernetes.io/part-of: marblerun
     app.kubernetes.io/version: {{ .Values.coordinator.version }}
 spec:
   type: ClusterIP
@@ -152,7 +152,7 @@ metadata:
     app.kubernetes.io/component: mesh-api
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator-mesh-api
-    app.kubernetes.io/part-of: {{ .Release.Namespace }}
+    app.kubernetes.io/part-of: marblerun
     app.kubernetes.io/version: {{ .Values.coordinator.version }}
 spec:
   type: ClusterIP
--- a/charts/templates/coordinator.yaml
+++ b/charts/templates/coordinator.yaml
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/component: coordinator
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator
-    app.kubernetes.io/part-of: {{ .Release.Namespace }}
+    app.kubernetes.io/part-of: marblerun
     app.kubernetes.io/version: {{ .Values.coordinator.version }}
     {{ .Values.global.coordinatorComponentLabel }}: coordinator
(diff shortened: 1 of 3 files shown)


charts/templates/coordinator.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/component: coordinator
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator
-    app.kubernetes.io/part-of: marblerun
+    app.kubernetes.io/part-of: {{ .Release.Namespace }}
```
daniel-weisse, line 10:
```suggestion
    app.kubernetes.io/part-of: marblerun
```
There already is a label to filter by namespace (`{{ .Values.global.coordinatorNamespaceLabel }}: {{ .Release.Namespace }}`)
This label is supposed to allow application level filter, its not scoped to a namespace.

charts/templates/coordinator.yaml
```diff
@@ -131,10 +131,10 @@ metadata:
     app.kubernetes.io/component: client-api
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator-client-api
-    app.kubernetes.io/part-of: marblerun
+    app.kubernetes.io/part-of: {{ .Release.Namespace }}
     app.kubernetes.io/version: {{ .Values.coordinator.version }}
 spec:
-  type: ClusterIP
+  type: NodePort
```
daniel-weisse, line 137: Why this change?
There should be no namespace issues with ClusterIP

charts/templates/marble-injector.yaml
```diff
@@ -75,7 +75,7 @@ spec:
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: marble-injector
+  name: marble-injector-{{ .Release.Namespace }}
```
daniel-weisse, line 78: Actually had this part in my local changes, missed pushing it
Thanks for catching :+1:

daniel-weisse (changes requested):

daniel-weisse:
Changed base to my PR, should make review a bit easier

> Currently the user needs to create the namespace and create the secret manually

Is this regarding the current state on main, or the state of my PR?
Because my PR should already fix that issue by using cert-manager to provision it

OverOrion:
> Changed base to my PR, should make review a bit easier
> 
> > Currently the user needs to create the namespace and create the secret manually
> 
> Is this regarding the current state on main, or the state of my PR? Because my PR should already fix that issue by using cert-manager to provision it

We have written an install script before your PR and we were working on it at the same time, I just removed the manual secret creation step from the script and it works as well! :)

charts/templates/coordinator.yaml
OverOrion: You are right, nice catch!

charts/templates/coordinator.yaml
OverOrion: Sorry for that, it is a leftover.  While we were testing it it was easier for us to have that port externally available.

charts/templates/coordinator.yaml
```diff
@@ -23,7 +23,7 @@ spec:
         app.kubernetes.io/component: coordinator
         app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
         app.kubernetes.io/name: coordinator
-        app.kubernetes.io/part-of: marblerun
+        app.kubernetes.io/part-of: {{ .Release.Namespace }}
```
daniel-weisse, line 26:
```suggestion
        app.kubernetes.io/part-of: marblerun
```
These should also just be labeled `marblerun`

charts/templates/coordinator.yaml
```diff
@@ -110,7 +110,7 @@ metadata:
     app.kubernetes.io/component: persistent-storage
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator-pv-claim
-    app.kubernetes.io/part-of: marblerun
+    app.kubernetes.io/part-of: {{ .Release.Namespace }}
```
daniel-weisse, line 113:
```suggestion
    app.kubernetes.io/part-of: marblerun
```

charts/templates/coordinator.yaml
```diff
@@ -131,7 +131,7 @@ metadata:
     app.kubernetes.io/component: client-api
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator-client-api
-    app.kubernetes.io/part-of: marblerun
+    app.kubernetes.io/part-of: {{ .Release.Namespace }}
```
daniel-weisse, line 134:
```suggestion
    app.kubernetes.io/part-of: marblerun
```

charts/templates/coordinator.yaml
```diff
@@ -152,7 +152,7 @@ metadata:
     app.kubernetes.io/component: mesh-api
     app.kubernetes.io/created-by: {{ .Values.global.createdBy }}
     app.kubernetes.io/name: coordinator-mesh-api
-    app.kubernetes.io/part-of: marblerun
+    app.kubernetes.io/part-of: {{ .Release.Namespace }}
```
daniel-weisse, line 155:
```suggestion
    app.kubernetes.io/part-of: marblerun
```

charts/templates/coordinator.yaml
OverOrion: Thanks again, fixed in the latest fixup commit.

PR #376, opened 2023-02-24
Set all required resources keys for Intel Device Plugin

daniel-weisse:
### Proposed changes
- Set all 3 resource keys for intel SGX device plugin
- rework resource key checking logic

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -470,14 +470,36 @@ func errorAndCleanup(err error, kubeClient kubernetes.Interface) error {
 	return err
 }
 
+// needsDeletion checks if an existing key of a helm chart should be deleted.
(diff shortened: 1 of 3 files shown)


cli/cmd/install.go
```diff
@@ -470,5 +470,15 @@ func errorAndCleanup(err error, kubeClient kubernetes.Interface) error {
 	return err
 }
 
+func needsDeletion(existingKey, sgxKey string) bool {
```
m1ghtym0, line 473: I don't understand this function.
Doesn't that return true for all non-sgx related values?
Can you make a comment for the second check? -> Reads like we want to delete the `sgx.intel.com/epc` value if the intel plugin is detected?

m1ghtym0 (approved):

PR #380, opened 2023-03-06
Refactor CLI code

daniel-weisse:
### Proposed changes
- Remove deprecated `gramine-prepare` command
- Move CLI code to `cli/internal/cmd/`
- Move Coordinator REST interaction to separate package
- Move helm interaction to separate package
- Move Kubernetes client setup to separate package
  - Could be further refactored by only returning a limited interface instead of a fully fledged Kubernetes client, but I think its fine for now
-  Add package for file interaction


cli/internal/rest/rest.go
```diff
@@ -0,0 +1,326 @@
...
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		// return data field of JSON response
+		data := gjson.GetBytes(respBody, dataField).String()
+		return []byte(data), nil
+	case http.StatusUnauthorized:
+		return nil, fmt.Errorf("GET %s: unable to authorize user: %s", uri.String(), msg)
+	default:
+		return nil, fmt.Errorf("GET %s: %d %s %s", uri.String(), resp.StatusCode, http.StatusText(resp.StatusCode), msg)
```
thomasten, line 222:
```suggestion
		return nil, fmt.Errorf("GET %s: authorizing user: %s", uri.String(), msg)
	default:
		return nil, fmt.Errorf("GET %s: %s %s", uri.String(), resp.Status, msg)
```

cli/internal/cmd/manifestUpdate.go
```diff
@@ -0,0 +1,254 @@
...
+func runUpdateApply(cmd *cobra.Command, args []string) error {
+	manifestFile := args[0]
+	hostname := args[1]
+
+	client, err := rest.NewAuthenticatedClient(cmd, hostname)
+	if err != nil {
+		return err
+	}
+
+	// Load manifest
```
thomasten, line 123:
```suggestion
```
Superfluous. Name of the func is clear enough.

cli/internal/cmd/manifestVerify.go
```diff
@@ -66,24 +67,24 @@ func getSignatureFromString(manifest string) (string, error) {
...
 		return "", err
 	}
 	return cliManifestSignature(rawManifest), nil
 }
 
 // cliManifestVerify verifies if a signature returned by the MarbleRun Coordinator is equal to one locally created.
-func cliManifestVerify(out io.Writer, localSignature string, host string, cert []*pem.Block) error {
-	remoteSignature, err := cliDataGet(host, "manifest", "data.ManifestSignature", cert)
+func cliManifestVerify(cmd *cobra.Command, localSignature string, client getter) error {
+	res, err := client.Get(cmd.Context(), rest.ManifestEndpoint, http.NoBody)
```
thomasten, line 79:
```suggestion
	resp, err := client.Get(cmd.Context(), rest.ManifestEndpoint, http.NoBody)
```
for consistency

cli/internal/cmd/manifestVerify.go
```diff
@@ -66,24 +67,24 @@ func getSignatureFromString(manifest string) (string, error) {
...
 	if err != nil {
 		return err
 	}
-
-	if string(remoteSignature) != localSignature {
+	remoteSignature := gjson.GetBytes(res, "ManifestSignature").String()
+	if remoteSignature != localSignature {
 		return fmt.Errorf("remote signature differs from local signature: %s != %s", string(remoteSignature), localSignature)
```
thomasten, line 85:
```suggestion
		return fmt.Errorf("remote signature differs from local signature: %s != %s", remoteSignature, localSignature)
```

cli/internal/cmd/manifest_test.go
```diff
@@ -0,0 +1,469 @@
...
+			err := cliManifestUpdateAcknowledge(cmd, []byte("manifest"), tc.poster)
+			if tc.wantErr {
+				assert.Error(err)
+				return
+			}
+			assert.NoError(err)
+			assert.Contains(out.String(), "Acknowledgement successful")
+			assert.Contains(out.String(), string(tc.poster.response))
+			assert.Equal(tc.poster.requestPath, rest.UpdateStatusEndpoint)
+			assert.Equal(tc.poster.header, rest.ContentJSON)
```
thomasten, line 366:
```suggestion
			assert.Equal(rest.UpdateStatusEndpoint, tc.poster.requestPath)
			assert.Equal(rest.ContentJSON, tc.poster.header)
```

thomasten (commented):
I got one question in certificate_test.go. Otherwise only nits. Choose yourself which ones you want to apply.

cli/cmd/root.go
```diff
@@ -7,14 +7,23 @@
...
 )
 
 // Execute starts the CLI.
 func Execute() error {
-	return NewRootCmd().Execute()
+	cobra.EnableCommandSorting = false
+	rootCmd := NewRootCmd()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	return rootCmd.ExecuteContext(ctx)
```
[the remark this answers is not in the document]
daniel-weisse: I copied this mostly from our Constellation code, where we use the cancel for "soft" interrupts.
Since we do not use that here, we can probably just use `context.Background()`

cli/internal/cmd/certificate_test.go
```diff
@@ -0,0 +1,223 @@
...
+		},
+		"one cert": {
+			file: file.New("unit-test", afero.NewMemMapFs()),
+			certs: []*pem.Block{
+				{
+					Type:  "CERTIFICATE",
+					Bytes: []byte("ROOT CERTIFICATE"),
+				},
+			},
+		},
```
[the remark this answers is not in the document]
daniel-weisse: Just checked against the old implementation again.
We used to only write an intermediate certificate if we received 2 or more certificates, otherwise we print a warning and exit without error.
The current implementation writes a certificate if we receive 1 or more certificates, and errors otherwise.

IIRC this command was originally written around the time when we first introduced intermediate certificates for the Coordinator.
Since the Coordinator should always return 2 certificates (intermediate and root), I think the correct way of handling this would be:
* write intermediate Certificate if we receive 2 certificates
* error in all other cases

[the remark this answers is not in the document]
daniel-weisse: Decided to align with previous behavior, i.e. write file if we receive 2 or more, print a warning otherwise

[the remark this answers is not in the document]
daniel-weisse: Actually, nevermind. Lets make this command have consistent behavior with what is expected of  the Coordinator and throw an error if we don't receive 2 certificates.

cli/internal/rest/rest.go
```diff
@@ -0,0 +1,311 @@
...
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		// return data field of JSON response
+		data := gjson.GetBytes(respBody, dataField).String()
+		return []byte(data), nil
+	case http.StatusUnauthorized:
+		return nil, fmt.Errorf("GET %s: authorizing user: %s", req.URL.String(), msg)
+	default:
+		return nil, fmt.Errorf("GET %s: %s %s", req.URL.String(), resp.Status, msg)
```
thomasten, line 241:
```suggestion
		return nil, fmt.Errorf("%s %s: authorizing user: %s", req.Method, req.URL.String(), msg)
	default:
		return nil, fmt.Errorf("%s %s: %s %s", req.Method, req.URL.String(), resp.Status, msg)
```

thomasten (approved):

PR #382, opened 2023-03-15
Allow usage of custom PVC for Coordinator storage

daniel-weisse:
### Proposed changes
- Update helm chart to allow users the usage of pre-created PVCs for the Coordinator's persistent storage.

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -71,8 +71,12 @@ coordinator:
       sgx.intel.com/enclave: 1
       sgx.intel.com/provision: 1
 
+
   # Set the storage class for your cluster
-  # storageClass:
+  storageClass: ""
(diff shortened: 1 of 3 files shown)


charts/values.yaml
```diff
@@ -64,15 +77,21 @@ coordinator:
...
     limits:
-      # Disable this if your running on a cluster without an SGX Device Plugin
+      # Resource request to use Intel SGX Device Plugin
+      # If you are running in simulation mode, or are using a different plugin,
+      # update these values accordingly
       sgx.intel.com/epc: "10Mi"
       sgx.intel.com/enclave: 1
       sgx.intel.com/provision: 1
 
   # Set the storage class for your cluster
```
m1ghtym0, line 90: Would make sense to extend the comment similar to the one below for the PVC

PR #387, opened 2023-04-12
split util package to reduce dependencies of premain

thomasten:
The premain binary currently contains multiple k8s packages because it imports the util package. Split the package to get rid of them.


util/k8sutil/k8sutil.go
daniel-weisse, line 1: Missing license header

daniel-weisse (approved):

PR #388, opened 2023-04-17
Only work on store transactions

daniel-weisse:
### Proposed changes
- Stop embedding store Wrappers in the Coordinator parts
- Only read data from transactions to make sure all information obtained during one call comes from the same state
- Marble Activation: Only commit state if MaxActivations is something we care about, dismiss state changes otherwise.
  - This reduces the amount of state updates we need to perform once MarbleRun is properly set up and running.

### DIFF
--- a/coordinator/clientapi/clientapi_test.go
+++ b/coordinator/clientapi/clientapi_test.go
@@ -590,7 +590,11 @@ func (c *fakeCore) RequireState(states ...state.State) error {
 	return errors.New("core is not in expected state")
 }
 
-func (c *fakeCore) AdvanceState(newState state.State, _ store.Transaction) error {
+func (c *fakeCore) AdvanceState(newState state.State, _ interface {
+	PutState(state.State) error
+	GetState() (state.State, error)
+},
+) error {
 	if c.advanceStateErr != nil {
 		return c.advanceStateErr
 	}
--- a/coordinator/clientapi/clientapi_test.go
+++ b/coordinator/clientapi/clientapi_test.go
@@ -143,7 +143,7 @@ func TestGetCertQuote(t *testing.T) {
 				require.NoError(err)
 			}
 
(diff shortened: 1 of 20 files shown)


coordinator/store/wrapper/wrapper.go
```diff
@@ -22,6 +23,24 @@ import (
...
+		return Wrapper{}, nil, nil, err
+	}
+	wrapper = New(tx)
+	rollback = func() {
+		tx.Rollback()
+	}
+	commit = func(ctx context.Context) error {
+		return tx.Commit(ctx)
+	}
+	return wrapper, rollback, commit, nil
```
thomasten, line 41:
```suggestion
	return New(tx), tx.Rollback, tx.Commit, nil
```
AFAIK this is semantically equivalent

coordinator/clientapi/clientapi_test.go
```diff
@@ -164,51 +130,59 @@ func TestGetCertQuote(t *testing.T) {
...
+				txHandle: tc.store,
+				log:      log,
+			}
+
+			var intermediateCert, rootCert *x509.Certificate
+			if !tc.wantErr {
+				intermediateCert = testutil.GetCertificate(t, tc.store, constants.SKCoordinatorIntermediateCert)
+				require.NoError(err)
+				rootCert = testutil.GetCertificate(t, tc.store, constants.SKCoordinatorRootCert)
+				require.NoError(err)
```
thomasten, line 143:
```suggestion
				intermediateCert = testutil.GetCertificate(t, tc.store, constants.SKCoordinatorIntermediateCert)
				rootCert = testutil.GetCertificate(t, tc.store, constants.SKCoordinatorRootCert)
```
and some other places where you switched to using testutil

thomasten (approved):
LGTM!

PR #392, opened 2023-05-03
Move file saving from Sealer to StdStore

daniel-weisse:
### Proposed changes
- Refactor the `Sealer` used by the Coordinator for data encryption to not handle file saving itself
  - The existing logic heavily limited the usability of the interface, e.g. when implementing a custom store interface, one would most likely also have to implement a custom sealer if they wanted to persist the data differently
- Move file saving into `StdStore`

### DIFF
--- a/coordinator/seal/noenclavesealer.go
+++ b/coordinator/seal/noenclavesealer.go
@@ -38,7 +38,9 @@ func (s *NoEnclaveSealer) Unseal(sealedData []byte) ([]byte, []byte, error) {
 	decryptedData, err := ecrypto.Decrypt(cipherText, s.encryptionKey, nil)
 	if err != nil {
 		// NoEnclaveSealer needs to return EncryptionKeyError if decryption fails
-		// This is required for integration tests to work
+		// so that the Coordinator enters recovery mode.
+		// This is required for integration tests to work that corrupt the encryption key.
+		// (The AESGCMSealer already fails in UnsealEncryptionKey with this error.)
 		return unencryptedData, nil, fmt.Errorf("decrypting sealed data: %w", &EncryptionKeyError{Err: err})
 	}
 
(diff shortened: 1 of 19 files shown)


coordinator/seal/seal.go
```diff
@@ -4,337 +4,150 @@
...
-func (s *NoEnclaveSealer) getFname(basename string) string {
-	return filepath.Join(s.sealDir, basename)
-}
-
-func (s *NoEnclaveSealer) loadEncryptionKey() error {
-	if s.encryptionKey != nil {
-		return nil
+// sealData encrypts data and seals it with the given key.
+// It returns the encrypted data prefixed with the unencrypted data and it's length.
+func sealData(unencryptedData, toBeEncrypted, encryptionKey []byte) ([]byte, error) {
```
thomasten, line 134: I think we now can have unit tests for some func in this file

coordinator/seal/noenclavesealer.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if s.encryptionKey == nil {
+		return unencryptedData, nil, fmt.Errorf("decrypting sealed data: %w", ErrMissingEncryptionKey)
+	}
+
+	// Decrypt data with the unsealed encryption key and return it
+	decryptedData, err := ecrypto.Decrypt(cipherText, s.encryptionKey, nil)
+	if err != nil {
+		// NoEnclaveSealer needs to return EncryptionKeyError if decryption fails
+		// This is required for integration tests to work
```
thomasten, line 41:
```suggestion
		// NoEnclaveSealer needs to return EncryptionKeyError if decryption fails
		// so that the Coordinator enters recovery mode.
		// This is required for integration tests to work that corrupt the encryption key.
		// (The AESGCMSealer already fails in UnsealEncryptionKey with this error.)
```
correct?

coordinator/store/stdstore/stdstore.go
```diff
@@ -124,10 +159,25 @@ func (s *StdStore) SetRecoveryData(recoveryData []byte) {
...
+
+	encryptedKey, err := s.sealer.SetEncryptionKey(encryptionKey)
+	if err != nil {
+		return fmt.Errorf("encrypting data key: %w", err)
+	}
+
+	// Write the sealed encryption key to disk
+	if err = s.fs.WriteFile(filepath.Join(s.sealDir, SealedKeyFname), encryptedKey, 0o600); err != nil {
+		// Reset encryption key if writing to disk fails
+		_, _ = s.sealer.SetEncryptionKey(encryptionKey)
```
thomasten, line 173: This sets the new key again, i.e., it's a no-op.
I think we should factor out the respective code into a new func SealEncryptionKey and let SetEncryptionKey only do what it says.
And then do:
SealEncryptionKey()
WriteFile()
SetEncryptionKey()

thomasten (approved):

coordinator/seal/noenclavesealer.go
daniel-weisse: Moved the error out of the sealer interface, which makes this a bit cleaner

PR #397, opened 2023-05-30
fix: return error on SetManifest if state cannot be committed

daniel-weisse:
### Proposed changes
- Return an error to the user if committing the new state after `SetManifest` fails
  - Previously, this only printed an error in the Coordinator's log.
  - Previously, users received a success message, even though committing the state failed, and the Coordinator remained in the `AcceptingManifest` state
- Use `zaptest.NewLogger(t)` in unit tests to set up logging, instead of `zap.NewDevelopment()`

### DIFF
--- a/coordinator/core/core_test.go
+++ b/coordinator/core/core_test.go
@@ -116,7 +116,6 @@ func TestRecover(t *testing.T) {
 	require := require.New(t)
 	ctx := context.Background()
 
-	// setup mock zaplogger which can be passed to Core
 	zapLogger := zaptest.NewLogger(t)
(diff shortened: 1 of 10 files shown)


coordinator/core/core_test.go
```diff
@@ -120,9 +117,7 @@ func TestRecover(t *testing.T) {
 	ctx := context.Background()
 
 	// setup mock zaplogger which can be passed to Core
```
thomasten, line 119:
```suggestion
```

thomasten (approved):

PR #399, opened 2023-06-13
docs: add versions

m1ghtym0:
### Proposed changes
- Add versions to docusaurus theme
- Add version for 1.1.0


docs/versions.json
```diff
@@ -0,0 +1,3 @@
+[
+  "1.1.0"
```
thomasten, line 2: this should be just 1.1

thomasten (changes requested):

daniel-weisse (approved):
works

PR #433, opened 2023-06-28
cli: require chart path when using enterprise access token

thomasten:
### Proposed changes
- Enterprise chart isn't included in the edgeless helm repo anymore, so make --marblerun-chart-path required for enterprise.
- Propose to let the cli do the base64(token:token) encoding. Not sure if this was left to the user on purpose or not. ~I can drop that commit if you have concerns (other auth secret formats?). Or I can add a simple heuristic, e.g., starting with ghp_~ Actually, I prefer accepting both inputs and convert the PAT automatically for convenience, so I've just added the check. Should be fine now. (Let me know if not.)

### DIFF
--- a/cli/internal/cmd/install.go
+++ b/cli/internal/cmd/install.go
@@ -96,7 +96,7 @@ func cliInstall(cmd *cobra.Command, helmClient *helm.Client, kubeClient kubernet
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


daniel-weisse (approved):
Accepting both formats for PATs sounds like a good idea

PR #444, opened 2023-07-18
Remove dependency on unmaintained openssl lib

daniel-weisse:
### Proposed changes
Rewrite the "TestOpenSSLVerify" function to no longer use openssl, since the library we depended on is unmaintained and does not support newer OpenSSL versions.
The certificate chain is now verified using pure Go.


m1ghtym0 (approved):
As discussed, LGTM

PR #477, opened 2023-09-05
Remove unused mermaid npm module

daniel-weisse:
### Proposed changes
- We don't have any mermaid graphs in our docs, so we can safely remove the dependency
- Remove unused npm packages


daniel-weisse:
Is there some way to trigger a docs build on this PR?

PR #528, opened 2023-11-17
Allow adding additional IPs for Coordinator root cert

daniel-weisse:
### Proposed changes
- Allow adding additional IPs to the Coordinator root cert by correctly parsing them from the `EDG_COORDINATOR_DNS_NAMES` env variable

### Additional info
Previously, IP addresses were incorrectly added as additional DNS names of the certificate. This means any generated Certificate was only signed for the IP adresses `127.0.0.1` and the ipv6 equivalent. This effectively made it required to provide a DNS name when exposing the client API publicly, e.g. when running the Coordinator in Kubernetes and using a Loadbalancer to assign a public IP to the client API.

### DIFF
--- a/coordinator/crypto/crypto.go
+++ b/coordinator/crypto/crypto.go
@@ -15,7 +15,6 @@ import (
 	"crypto/x509/pkix"
 	"fmt"
 	"math"
-	"net"
 	"time"
 
 	"github.com/edgelesssys/marblerun/util"
@@ -24,7 +23,7 @@ import (
 // GenerateCert creates a new certificate with the given parameters.
 // If privk is nil, a new private key is generated.
 func GenerateCert(
-	dnsNames []string, commonName string, privk *ecdsa.PrivateKey,
+	subjAltNames []string, commonName string, privk *ecdsa.PrivateKey,
 	parentCertificate *x509.Certificate, parentPrivateKey *ecdsa.PrivateKey,
 ) (*x509.Certificate, *ecdsa.PrivateKey, error) {
 	// Generate private key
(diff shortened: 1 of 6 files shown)


coordinator/crypto/crypto.go
```diff
@@ -44,13 +45,23 @@ func GenerateCert(
 		return nil, nil, fmt.Errorf("generating serial number: %w", err)
 	}
 
+	var additionalDNSNames []string
+	var additionalIPs []net.IP
+	for _, name := range dnsNames {
```
thomasten, line 50:
```suggestion
	var dnsNames []string
	var additionalIPs []net.IP
	for _, name := range subjAltNames {
```
nit: these names would be more precise

coordinator/crypto/crypto.go
```diff
@@ -44,13 +45,23 @@ func GenerateCert(
 		return nil, nil, fmt.Errorf("generating serial number: %w", err)
 	}
 
+	var additionalDNSNames []string
```
thomasten, line 48: We should have a unit test for this func. For now it would be ok if you only add cases that cover your change.

coordinator/crypto/crypto.go
```diff
@@ -44,13 +45,23 @@ func GenerateCert(
...
+
 	template := x509.Certificate{
 		SerialNumber: serialNumber,
 		Subject: pkix.Name{
 			CommonName: commonName,
 		},
-		DNSNames:    dnsNames,
-		IPAddresses: util.DefaultCertificateIPAddresses,
+		DNSNames:    additionalDNSNames,
+		IPAddresses: append(util.DefaultCertificateIPAddresses, additionalIPs...),
```
thomasten, line 64: should we support this for MARBLE_DNS_NAMES too?

thomasten, line 64: update docs to say "DNS names and IP addresses"?

thomasten (approved):

PR #607, opened 2024-03-12
deps: Update EGo, ERT and Go versions

daniel-weisse:
### Proposed changes
- Update EGo to `v1.5.0`
- Update Go to `v1.21.8`
- Update ERT to `v0.4.3`


.github/workflows/lint.yml
```diff
@@ -25,6 +25,9 @@ jobs:
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
+      - run: |
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
```
thomasten, line 29: Do you know why this is needed now and wasn't needed before?

thomasten (approved):

.github/workflows/lint.yml
daniel-weisse: I'm not 100% sure, but golangci-lint with Go 1.21 seems to have issues with `buildvcs` otherwise.
We already use the same git command to mark the repository safe for the unittest workflow, but didnt here, since so far golangci-lint didnt complain

PR #627, opened 2024-04-16
Retire usage of era in CLI functions

daniel-weisse:
### Proposed changes
- Adapt [era](https://github.com/edgelesssys/era) code and move it to MarbleRun directly instead of having it as external dependency
- Fix log message when fetching latest era config from GitHub
- Use better error message when trying to verify a manifest on a Coordinator which has no manifest set (e.g. is in recovery mode / waiting for manifest)

### DIFF
--- a/cli/internal/cmd/manifestVerify.go
+++ b/cli/internal/cmd/manifestVerify.go
@@ -99,6 +99,11 @@ func cliManifestVerify(cmd *cobra.Command, localSignature string, client getter)
 		return err
 	}
 	remoteSignature := gjson.GetBytes(resp, "ManifestSignature").String()
+
+	if remoteSignature == "" {
+		return errors.New("Coordinator returned no manifest signature. Is the Coordinator in the correct state?")
+	}
+
 	if remoteSignature != localSignature {
 		return fmt.Errorf("remote signature differs from local signature: %s != %s", remoteSignature, localSignature)
 	}
--- a/internal/attestation/attestation.go
+++ b/internal/attestation/attestation.go
@@ -1,0 +1,203 @@
+// Copyright (c) Edgeless Systems GmbH.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+package attestation
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"crypto/tls"
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/edgelesssys/ego/attestation"
+	"github.com/edgelesssys/ego/attestation/tcbstatus"
+	"github.com/edgelesssys/ego/eclient"
+	"github.com/tidwall/gjson"
+)
+
+// Config is the expected attestation metadata of a MarbleRun Coordinator enclave.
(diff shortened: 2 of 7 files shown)


cli/internal/cmd/manifestVerify.go
```diff
@@ -99,6 +99,11 @@ func cliManifestVerify(cmd *cobra.Command, localSignature string, client getter)
 		return err
 	}
 	remoteSignature := gjson.GetBytes(resp, "ManifestSignature").String()
+
+	if remoteSignature == "" {
+		return errors.New("Coordinator returned no manifest signature. Is the Coordinator in the correct state?")
```
thomasten, line 104: shouldn't the coordinator return an error instead?

internal/attestation/attestation.go
```diff
@@ -0,0 +1,203 @@
...
+
+// httpGetCertQuote requests the Coordinator's quote and certificate chain.
+func httpGetCertQuote(ctx context.Context, host string, nonce []byte) (string, []byte, error) {
+	client := http.Client{
+		Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}},
+	}
+
+	url := url.URL{Scheme: "https", Host: host, Path: "quote"}
+	if nonce != nil {
+		url.Query().Add("nonce", hex.EncodeToString(nonce))
```
thomasten, line 166:
```suggestion
	if len(nonce) > 0 {
		url.Query().Add("nonce", base64.URLEncoding.EncodeToString(nonce))
```

internal/attestation/attestation.go
```diff
@@ -0,0 +1,203 @@
...
+			return "", nil, errors.New(resp.Status + ": " + errorMessage.String())
+		}
+		return "", nil, errors.New(resp.Status + ": " + string(body))
+	}
+
+	var certQuote certQuoteResp
+	if err := json.Unmarshal([]byte(gjson.GetBytes(body, "data").String()), &certQuote); err != nil {
+		return "", nil, err
+	}
+	resp.Body.Close()
```
thomasten, line 196:
```suggestion
```

internal/attestation/attestation_test.go
```diff
@@ -0,0 +1,317 @@
...
+				assert.Equal(tc.tcbStatus, tcbStatus)
+			} else if tc.wantErr {
+				assert.Error(err)
+				assert.NotErrorIs(err, attestation.ErrTCBLevelInvalid)
+				return
+			} else {
+				require.NoError(err)
+			}
+
+			assert.EqualValues(expectedCert, pem.EncodeToMemory(actualCerts[0]))
```
thomasten, line 225:
```suggestion
			assert.EqualValues(expectedCert, pem.EncodeToMemory(actualCerts[0]))
			assert.Equal(quote, actualQuote)
```

cli/internal/cmd/manifestVerify.go
daniel-weisse: Currently it does not.
Though we may also just add a `RequireState` block to `GetManifest`

thomasten: Do you remember if there is a reason that GetManifestSignature doesn't return the error?
Anyway, your fix is an improvement to the status quo, so I leave it up to you whether to change the Coordinator's behavior or to check for emptiness here.

thomasten (approved):

cli/internal/cmd/manifestVerify.go
daniel-weisse: I don't think there ever was a "good" reason, the client API was just never built with the case of no manifest being set when calling this endpoint in mind.
Since changing this would _technically_ break API compatibility, I would keep it as is on the server side for now.

PR #644, opened 2024-05-03
cli: allow users to use a custom nonce for SGX quote verification using `--nonce` flag

daniel-weisse:
### Proposed changes
- Add a new flag to the CLI: `--nonce`
  - Takes a string as argument which, if set, is used by the Coordinator to generate to generate an SGX quote
    - If not set the Coordinator creates a quote over `sha256(rootCert)` (default and current behvaior)
    - If set the Coordinator creates a quote over `sha256(rootCert+nonce)`
    - The Coordinator uses the literal byte value of the (ASCII) string nonce. In the future we may add options to supply a a nonce hex or base64 encoded to allow for a wider range of nonce values
- Add a new endpoint to the Coordinators REST API: `/api/v2/quote`
  - This endpoint behaves the same as `/quote`, but additionally allows sending a nonce as part of the request query: 
  
    ```
    https://<host>:<port>/api/v2/quote?nonce=<base64-url-encoded-nonce>
    ````
  - If a nonce is set, the Coordinator will generate a new quote over said nonce instead of using the stored quote generated at start up

- Fix for the CI generating docs which claimed the default Coordinator cert path as `/github/home/.config/marblerun/coordinator-cert.pem`
  - The CI now generates docs which use `$HOME/.config` as a placeholder for the users actual config dir

### Additional info
- AB#4030


coordinator/clientapi/clientapi.go
```diff
@@ -139,8 +139,18 @@ func (a *ClientAPI) GetCertQuote(ctx context.Context) (cert string, certQuote []
 		return "", nil, errors.New("pem.EncodeToMemory failed for intermediate certificate")
 	}
 
+	// Get existing quote for root cert, or generate a new one over provided nonce
+	var reportData []byte
+	if nonce != nil {
```
thomasten, line 144:
```suggestion
	if len(nonce) > 0 {
```
In most cases, don't treat nil and empty slices differently. (See also https://github.com/edgelesssys/wiki/blob/master/company/gostyle.md#nil-is-a-valid-slice)

coordinator/clientapi/clientapi_test.go
```diff
@@ -120,6 +121,14 @@ func TestGetCertQuote(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		"get quote with nonce": {
+			store: prepareDefaultStore(),
+			core: &fakeCore{
+				state: state.AcceptingMarbles,
+				quote: []byte("quote"),
+			},
+			nonce: []byte("nonce"),
```
thomasten, line 130: for a meaningful test, the fakeCore should do something with the nonce and we should have wantQuote

coordinator/core/core.go
```diff
@@ -278,8 +278,17 @@ func (c *Core) GetTLSMarbleRootCertificate(clientHello *tls.ClientHelloInfo) (*t
...
-	return c.quote
+// If reportData is not nil, a new quote is generated over the data and returned.
+func (c *Core) GetQuote(reportData []byte) ([]byte, error) {
+	quote := c.quote
+	if reportData != nil {
+		var err error
+		quote, err = c.qi.Issue(reportData)
+		if err != nil && err.Error() != "OE_UNSUPPORTED" {
+			return nil, QuoteError{err}
+		}
```
thomasten, line 289:
```suggestion
// If reportData is not empty, a new quote is generated over the data and returned.
func (c *Core) GetQuote(reportData []byte) ([]byte, error) {
	if len(reportData) == 0 {
		return c.quote, nil
	}
	quote, err := c.qi.Issue(reportData)
	if err != nil && err.Error() != "OE_UNSUPPORTED" {
		return nil, QuoteError{err}
```

coordinator/core/core.go
```diff
@@ -278,8 +278,17 @@ func (c *Core) GetTLSMarbleRootCertificate(clientHello *tls.ClientHelloInfo) (*t
 }
 
 // GetQuote returns the quote of the Coordinator.
-func (c *Core) GetQuote() []byte {
-	return c.quote
+// If reportData is not nil, a new quote is generated over the data and returned.
+func (c *Core) GetQuote(reportData []byte) ([]byte, error) {
```
thomasten, line 282: consider adding a test for this func

internal/attestation/attestation.go
```diff
@@ -162,12 +162,15 @@ func httpGetCertQuote(ctx context.Context, host string, nonce []byte) (string, [
 		Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}},
 	}
 
-	url := url.URL{Scheme: "https", Host: host, Path: "quote"}
+	path := "quote"
```
thomasten, line 165:
```suggestion
	path := "quote" // use v1 path for compatibility with old Coordinators if nonce is not provided
```

thomasten (approved):

PR #658, opened 2024-05-16
api: Implement a public Go API

daniel-weisse:
### Proposed changes
- Implement a public Go API
  - Provides an easy native Go interface for performing CLI commands that interact with the MarbleRun Coordinator
- Use Go API as a backend for CLI commands
  - This removes the `/cli/internal/rest` package, as http calls are made by the API
    - Cert loading and writing from the `rest` package was moved to the new `/cli/internal/certcache` package
- Heavily simplify the code in `/internal/attestation` and move it to `/api/attestation`
- Add `/api/v2/recover` endpoint with better machine readable responses

### Additional info
- AB#4009
- I would like to eventually also move the integration test to use the Go API, but as this PR is already quite a bit larger than I originally intended, this will be postponed to a different PR


api/api.go
```diff
@@ -0,0 +1,527 @@
+// Copyright (c) Edgeless Systems GmbH.
```
thomasten, line 1: agree that we should use this in the integration test (in an upcoming PR)

api/api.go
```diff
@@ -0,0 +1,527 @@
...
+
+func (v *VerifyOptions) setDefaults() {
+	if len(v.AcceptedTCBStatuses) == 0 {
+		v.AcceptedTCBStatuses = []string{"UpToDate", "SWHardeningNeeded"}
+	}
+}
+
+func verifyOptionsFromConfig(fs afero.Afero, configPath string) (VerifyOptions, error) {
+	var opts VerifyOptions
+	optsRaw, err := fs.ReadFile(configPath)
```
thomasten, line 428: did you plan to write unit tests, or what is the split func and use of afero about?

api/attestation/attestation.go
```diff
@@ -0,0 +1,124 @@
...
+	ProductID           uint16
+	Debug               bool
+	Nonce               []byte
+	AcceptedTCBStatuses []string
+}
+
+// VerifyCertificate verifies the Coordinator's TLS certificate against the Coordinator's SGX quote.
+// A config with the expected attestation metadata must be provided.
+// It returns the TCB status of the enclave, the quote, and an error, if any.
+func VerifyCertificate(certificate *pem.Block, quote []byte, config Config) error {
```
thomasten, line 55: This is crucial. We should have unit tests like internal/attestation/attestation_test.go

api/attestation/attestation.go
```diff
@@ -0,0 +1,124 @@
...
+		return err
+	}
+
+	if err := verifyReport(report, certificate.Bytes, config); err != nil {
+		return err
+	}
+
+	// Check if the TCB status is accepted
+	if !slices.Contains(config.AcceptedTCBStatuses, report.TCBStatus.String()) {
+		return NewTCBStatusError(report.TCBStatus)
```
thomasten, line 67: This is a behavior change for the CLI: UpToDate was always accepted, even if not in AcceptedTCBStatuses.
Also, the messages "TCB level accepted by configuration" and "Warning: TCB level invalid, but accepted by configuration" are gone.

cli/internal/certcache/cert.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, nil, errors.New("could not parse Coordinator root certificate from PEM data")
+	}
+	intermediate, err = x509.ParseCertificate(intermediatePEM.Bytes)
+	if err != nil {
+		return nil, nil, fmt.Errorf("parsing Coordinator intermediate certificate: %w", err)
+	}
+	root, err = x509.ParseCertificate(rootPEM.Bytes)
+	if err != nil {
+		return nil, nil, fmt.Errorf("parsing Coordinator root certificate: %w", err)
+	}
```
thomasten, line 103: nit: similar code in api.go - can this be moved to a helper func?

cli/internal/cmd/cmd.go
```diff
@@ -9,62 +9,131 @@ package cmd
...
+	eraCfgRaw, err := os.ReadFile(eraConfig)
 	if err != nil {
-		return rest.VerifyCoordinatorOptions{}, err
+		return api.VerifyOptions{}, "", fmt.Errorf("reading era config file: %w", err)
+	}
+
+	var verifyOptions api.VerifyOptions
+	if err := json.Unmarshal(eraCfgRaw, &verifyOptions); err != nil {
+		return api.VerifyOptions{}, "", fmt.Errorf("unmarshalling era config: %w", err)
+	}
```
thomasten, line 83: can we use VerifyOptionsFromConfig here?

internal/attestation/attestation.go
```diff
@@ -89,7 +89,7 @@ func getCertificate(ctx context.Context, host string, nonce []byte, config Confi
 	}
 
 	report, verifyErr := verifyRemoteReport(quote)
-	if verifyErr != nil && verifyErr != attestation.ErrTCBLevelInvalid {
+	if verifyErr != nil && !errors.Is(verifyErr, attestation.ErrTCBLevelInvalid) {
```
thomasten, line 92: this package is not needed anymore

api/api.go
daniel-weisse: Yes, though I'm currently unsure how worthwhile they are for the functions here, which mostly just unmarshal and return a value.
I feel this is better handled by integration or e2e tests.
We may add unit tests for completeness nonetheless.

api/attestation/attestation.go
daniel-weisse: I could re-add usage of the `/internal/tcb` package here.
Though I would prefer to not print anything when just using the API outside of the CLI.
Maybe a global `SetLogger` function could solve the issue.
By default nothing is logged, and the CLI can use the function to instruct the API to log this information to stdout

api/api.go
thomasten: I agree that unit tests don't add much value here. We should better focus on integration/e2e and make sure that those cover all of the API.

daniel-weisse: Our e2e tests already cover usage of the API calls through the CLI, added a ticket to update the integration test: 
https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/4185

thomasten (approved):

PR #659, opened 2024-05-23
Add endpoint to verify and sign SGX quotes

daniel-weisse:
### Proposed changes
- Add a new endpoint `/api/v2/sign-quote` which verifies a given SGX quote and signs the quote + resulting TCB status using the Coordinator's root key
- Add functions to the `api` package to sign a quote and verify it


coordinator/clientapi/clientapi.go
```diff
@@ -741,3 +758,41 @@ func (a *ClientAPI) WriteSecrets(ctx context.Context, rawSecretManifest []byte,
 
 	return commit(ctx)
 }
+
+// SignQuote verifies the quote and signs it with the Coordinator's root key.
+func (a *ClientAPI) SignQuote(ctx context.Context, quote []byte) (signedQuote []byte, tcbStatus string, err error) {
```
thomasten, line 763:
```suggestion
func (a *ClientAPI) SignQuote(ctx context.Context, quote []byte) (signature []byte, tcbStatus string, err error) {
```

coordinator/clientapi/clientapi.go
```diff
@@ -741,3 +758,41 @@ func (a *ClientAPI) WriteSecrets(ctx context.Context, rawSecretManifest []byte,
...
+			a.log.Error("SignQuote failed", zap.Error(err))
+		} else {
+			a.log.Info("SignQuote successful")
+		}
+	}()
+
+	// Verify the quote
+	var verifyErr error
+	report, err := enclave.VerifyRemoteReport(quote)
+	if err != nil {
```
thomasten, line 776:
```suggestion
	report, err := enclave.VerifyRemoteReport(quote)
	if err != nil && !errors.Is(err, attestation.ErrTCBLevelInvalid) {
```
also consider adding a unit test for this func

coordinator/clientapi/clientapi.go
```diff
@@ -741,3 +758,41 @@ func (a *ClientAPI) WriteSecrets(ctx context.Context, rawSecretManifest []byte,
...
+		if err != nil {
+			a.log.Error("SignQuote failed", zap.Error(err))
+		} else {
+			a.log.Info("SignQuote successful")
+		}
+	}()
+
+	// Verify the quote
+	var verifyErr error
+	report, err := enclave.VerifyRemoteReport(quote)
```
thomasten, line 775: This API should accept both OE reports and raw SGX quotes. For the difference, see
https://github.com/edgelesssys/marblerun/blob/3e62cf3e2976ef09cdb3944edd719b61ac733f1b/marble/premain/util.go#L11-L17

coordinator/server/api.go
```diff
@@ -50,6 +50,21 @@ type ManifestSignatureResp struct {
 	Manifest []byte
 }
 
+// QuoteSignReq contains an SGX Quote to be verified and signed by the Coordinator.
+type QuoteSignReq struct {
+	// SGXQuote is the raw SGX quote data.
+	SGXQuote []byte `json:"sgxQuote"`
+}
+
+// QuoteSignResp contains the SGX Quote signed by the Coordinator using its root ECDSA key,
```
thomasten, line 59:
```suggestion
// QuoteSignResp contains the SGX Quote signature created by the Coordinator using its root ECDSA key,
```

coordinator/server/api.go
```diff
@@ -50,6 +50,21 @@ type ManifestSignatureResp struct {
...
+// as well as the TCB status of the Quote.
+type QuoteSignResp struct {
+	// TCBStatus is the TCB status of the SGX Quote.
+	TCBStatus string `json:"tcbStatus"`
+	// VerificationSignature is a signature over sha256(base64(SGXQuote)|TCBStatus|string(VerifySuccess)) signed by the root ECDSA key.
```
thomasten, line 64:
```suggestion
	// VerificationSignature is a signature over sha256(base64(SGXQuote)|TCBStatus) signed by the root ECDSA key.
```

coordinator/server/v2.go
```diff
@@ -70,10 +75,38 @@ func (s *clientAPIServerV2) recoverPost(w http.ResponseWriter, r *http.Request)
 	})
 }
 
-// writeJSONFailure wires a JSend failure response to the client.
-func writeJSONFailure(w http.ResponseWriter, v interface{}, httpErrorCode int) {
+// signQuotePost receives an SGX quote and returns a signed quote.
```
thomasten, line 78:
```suggestion
// signQuotePost receives an SGX quote and returns a signature for it.
```

coordinator/server/v2.go
```diff
@@ -70,10 +75,38 @@ func (s *clientAPIServerV2) recoverPost(w http.ResponseWriter, r *http.Request)
...
+// The Coordinator will verify the quote and sign it together with the TCB status of the quote using the root ECDSA key.
+func (s *clientAPIServerV2) signQuotePost(w http.ResponseWriter, r *http.Request) {
```
thomasten, line 80: This endpoint/feature should be opt-in via the manifest.

thomasten (changes requested):

coordinator/server/v2.go
```diff
@@ -70,10 +76,43 @@ func (s *clientAPIServerV2) recoverPost(w http.ResponseWriter, r *http.Request)
...
 
-// writeJSONFailure wires a JSend failure response to the client.
-func writeJSONFailure(w http.ResponseWriter, v interface{}, httpErrorCode int) {
+// signQuotePost receives an SGX quote and returns a signature for it.
+// The Coordinator will verify the quote and sign it together with the TCB status of the quote using the root ECDSA key.
+func (s *clientAPIServerV2) signQuotePost(w http.ResponseWriter, r *http.Request) {
+	// Check if the current manifest allows signing quotes
+	if !s.api.FeatureEnabled(r.Context(), manifest.FeatureSignQuoteEndpoint) {
+		writeJSONError(w, "sign-quote endpoint is not enabled", http.StatusForbidden)
+	}
```
thomasten, line 85:
```suggestion
		writeJSONError(w, "sign-quote endpoint is not enabled", http.StatusForbidden)
		return
	}
```

util/util.go
```diff
@@ -130,3 +132,18 @@ func CoordinatorCertChainFromPEM(pemChain []byte) (rootCert, intermediateCert *x
...
+// AddOEQuoteHeader adds an OpenEnclave quote header to the given quote.
+// If the quote already has a header, this is a no-op.
+func AddOEQuoteHeader(quote []byte) []byte {
+	quoteHeader := make([]byte, 16)
+	binary.LittleEndian.PutUint32(quoteHeader, 1)     // version
+	binary.LittleEndian.PutUint32(quoteHeader[4:], 2) // OE_REPORT_TYPE_SGX_REMOTE
+	binary.LittleEndian.PutUint64(quoteHeader[8:], uint64(len(quote)))
+
+	// If the quote already has a header, return it as is.
+	if len(quote) > 8 && slices.Equal(quoteHeader[:8], quote[:8]) {
```
thomasten, line 145: OE has at least two supported headers, so this will break if it's the other one. Checking whether the quote is a raw SGX quote seems to be the more correct way. However, SGX quotes also have multiple versions. So we can either check for the version known to us. Or we just try to verify the quote as is and if it fails, we try again with added header. A bit ugly, but probably stable also for future changes to OE or SGX quote versions.

thomasten (changes requested):

util/util.go
```diff
@@ -130,3 +131,53 @@ func CoordinatorCertChainFromPEM(pemChain []byte) (rootCert, intermediateCert *x
...
+	// Version must be 3
+	if binary.LittleEndian.Uint16(quote[:2]) != 3 {
+		return false
+	}
+
+	// Attestation Key Type of the Quote Header must be:
+	//  - 2: ECDSA-256 with P-256 curve
+	//  - 3: ECDSA-384 with P-384 curve
+	// 0 and 1 are reserved, anything higher is not valid
+	if binary.LittleEndian.Uint16(quote[2:4]) > 4 {
```
thomasten, line 163:
```suggestion
	if binary.LittleEndian.Uint16(quote[2:4]) > 3 {
```

util/util.go
```diff
@@ -130,3 +131,53 @@ func CoordinatorCertChainFromPEM(pemChain []byte) (rootCert, intermediateCert *x
...
+	if binary.LittleEndian.Uint32(quote[4:8]) != 0 {
+		return false
+	}
+
+	// QE SVN must be higher than 0
+	if binary.LittleEndian.Uint16(quote[8:10]) == 0 {
+		return false
+	}
+
+	// QE Vendor ID must be higher than 0
```
thomasten, line 177:
```suggestion
	// PCE SVN must be higher than 0
```

util/util_test.go
```diff
@@ -67,3 +69,84 @@ func TestXORBytes(t *testing.T) {
...
+			}(),
+		},
+		"invalid QE SVN": {
+			quote: func() []byte {
+				quote := append(sgxQuoteHeader, bytes.Repeat([]byte{0x0}, 128)...)
+				binary.LittleEndian.PutUint16(quote[8:10], 0)
+				return quote
+			}(),
+		},
+		"invalid QE vendor ID": {
```
thomasten, line 135:
```suggestion
		"invalid PCE SVN": {
```

thomasten (approved):
This feature should have an integration or e2e test. In case of the former, it could be added when changing the test to use the API.

PR #661, opened 2024-06-04
coordinator: update all existing endpoints to `/api/v2`

daniel-weisse:
### Proposed changes
- Update all existing endpoints to support `/api/v2`.

  This API:
  - Uses JSend fail messages in case the supplied input was incorrect
  - Wraps inputs and outputs in JSON structs. e.g. instead of just passing the manifest directly as JSON input, a struct `{manifest: <raw-manifest>}` is used.
  - Logs a deprecation warning if api v1 (e.g. ´/manifest` instead of `/api/v2/manifest`) is used

- Default to `/api/v2` when using the Go API.

### Additional info
- AB#4137


api/api.go
```diff
@@ -57,13 +55,20 @@ func VerifyCoordinator(ctx context.Context, endpoint string, opts VerifyOptions)
...
 
 	var args []string
-	path := rest.QuoteEndpoint
 	if len(opts.Nonce) > 0 {
 		nonce := base64.URLEncoding.EncodeToString(opts.Nonce)
 		args = []string{"nonce", nonce}
-		path = rest.V2API + rest.QuoteEndpoint // Nonce is only supported in v2 API
+
+	}
+	resp, err := client.Get(ctx, rest.V2API+rest.QuoteEndpoint, http.NoBody, args...)
```
thomasten, line 63:
```suggestion
	}

	resp, err := client.Get(ctx, rest.V2API+rest.QuoteEndpoint, http.NoBody, args...)
```

coordinator/clientapi/clientapi.go
```diff
@@ -177,30 +177,29 @@ func (a *ClientAPI) GetCertQuote(ctx context.Context, nonce []byte) (cert string
 // GetManifestSignature returns the hash of the manifest.
 //
 // Returns ECDSA signature, SHA256 hash and byte encoded representation of the active manifest.
```
thomasten, line 179:
```suggestion
// Returns ECDSA signature and byte encoded representation of the active manifest.
```

thomasten (approved):

PR #669, opened 2024-06-12
docs: fix search

thomasten:
Search plugin doesn't work with docusaurus 3.4.0. It generates wrong URLs. Downgrade until cmfcmf/docusaurus-search-local#205 is resolved.


daniel-weisse (approved):
thanks

PR #677, opened 2024-06-17
coordinator: add manifest option to seal with unique key or disable sealing

thomasten:
Currently, the Coordinator uses the product key to seal the data. Users may want to use the unique key instead or disable persistent storage altogether for ephemeral deployments.

### Proposed changes
- Add `Config.SealMode` field to the manifest
- Use seal key as configured in manifest


coordinator/store/stdstore/stdstore.go
```diff
@@ -199,35 +203,10 @@ func (s *StdStore) commit(data map[string][]byte) error {
 
 	s.mux.Lock()
 	defer s.mux.Unlock()
-	if !s.recoveryMode {
+	if !s.recoveryMode && s.sealMode != seal.ModeDisabled {
 		sealedData, err := s.sealer.Seal(s.recoveryData, dataRaw)
 		if err != nil {
-			if !errors.Is(err, seal.ErrMissingEncryptionKey) {
```
daniel-weisse, line 205: This introduces a change where we no longer write data to disk before a manifest is set.
I prefer the new behavior, but we might want to mention this in the changelog

daniel-weisse (approved):

daniel-weisse (approved):

PR #718, opened 2024-09-02
fix: don't sign quotes including added OE header

daniel-weisse:
There currently is an issue where the signature returned by the `/sign-quote` endpoint is generated over a quote including an OE quote header, even if the original quote does not include one, i.e. because it was generated by Gramine or Occlum.

### Proposed changes
- Use the raw given quote to generate the signature when calling `/sign-quote`


coordinator/clientapi/clientapi.go
```diff
@@ -785,12 +785,13 @@ func (a *ClientAPI) verifyAndSignQuote(
 ) (signature []byte, tcbStatus string, err error) {
 	// Add OE quote header if it is not present
 	// e.g. we are dealing with a raw SGX quote generated by Gramine or Occlum
+	oeQuote := quote
 	if util.IsRawSGXQuote(quote) {
-		quote = util.AddOEQuoteHeader(quote)
+		oeQuote = util.AddOEQuoteHeader(quote)
```
thomasten, line 790: Please extend TestSignQuote to cover this. (A crafted quote header satisfying IsRawSGXQuote should be enough.)

thomasten (approved):

PR #730, opened 2024-09-18
Derive marble private secrets using marble type in addition to UUID

daniel-weisse:
Currently, when a marble requests a private secret, that secret is derived from the coordinators root key using HKDF with the Marble's UUID and the name of the secret being used as salt for the derivation.
This means that any two marbles with the same UUID will receive the same secrets, even if they have different marble types.
Since this behavior is not always clear to users, and they should use shared secrets to facilitate this secret instead, this PR updates the secret generation code to include the name of the marble as Info parameter for the HKDF function so any two Marbles of different types will receive different secrets even if they have the same UUID.

### Proposed changes
- Include the Marble's Type in secret derivation along the secret name and Marble UUID
  -  the old behavior can be restored by setting the new `DisableSecretBinding` on a Marble

### DIFF
--- a/coordinator/manifest/manifest.go
+++ b/coordinator/manifest/manifest.go
@@ -84,6 +84,9 @@ type Marble struct {
 	Parameters Parameters
 	// TLS holds a list of tags which are specified in the manifest
 	TLS []string
+	// DisableSecretBinding causes the Coordinator to not include the Marble's name for secret derivation.
+	// Effectively, this enforces the same behavior of the Coordinator previous to version 1.6.0.
+	DisableSecretBinding bool
 }
 
 // Equal returns true if two Marble definitions are equal.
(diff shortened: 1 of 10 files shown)


coordinator/manifest/manifest.go
```diff
@@ -84,6 +84,9 @@ type Marble struct {
 	Parameters Parameters
 	// TLS holds a list of tags which are specified in the manifest
 	TLS []string
+	// DisableSecretBinding causes the Coordinator to not include the Marble's name for secret derivation.
+	// Effectively, this enforces the same behavior of the Coordinator previous to version 1.6.0.
+	DisableSecretBinding bool
```
thomasten, line 89: add documentation?

thomasten (approved):

PR #766, opened 2024-11-19
premain: enable JSON formatted logs if `EDG_LOG_FORMAT=json`

daniel-weisse:
### Proposed changes
- Fix two log messages of the marble-injector being incorrectly written at info level instead of error level
- Enable JSON formatted logging for MarbleRun PreMain in case the `EDG_LOG_FORMAT` env variable is set to `json`
  - To ensure all log statements issued by the PreMain are JSON formatted, if enabled, changes to EGo and ERT are also required. ERT logs non JSON messages during bootup, e.g. [here](https://github.com/edgelesssys/edgelessrt/blob/54be13676da027b25dda90a52e4a52b2fe2f40e6/src/tools/erthost/erthost.cpp#L98), and EGo does the same

### Additional info
- AB#4868

### DIFF
--- a/docs/docs/workflows/add-service.md
+++ b/docs/docs/workflows/add-service.md
@@ -5,6 +5,7 @@ Adding a service to your application requires three steps, which are described i
 ## **Step 1:** Get your service ready for MarbleRun
 
 To get your service ready for MarbleRun, you need to rebuild it with one of the supported [runtimes](../features/runtimes.md):
+
 * [EGo](../building-marbles/ego.md)
 * [Edgeless RT](https://github.com/edgelesssys/marblerun/blob/master/samples/helloc%2B%2B)
 * [Gramine](../building-marbles/gramine.md)
@@ -29,6 +30,7 @@ Now that your service is ready, you need to make two types of entries in the man
 As is described in more detail in the [writing a manifest hands-on](../workflows/define-manifest.md#packages), the manifest contains a section `Packages`, in which allowed enclave software packages are defined.
 
 #### EGo / EdgelessRT
+
 To add an entry for your EGo / EdgelessRT service, run the `oesign` tool on the enclave file you built in the previous step as follows. (`oesign` is installed with [Edgeless RT](https://github.com/edgelesssys/edgelessrt).)
 
 ```bash
@@ -49,7 +51,6 @@ The tool's output is similar to the following.
 #### Gramine
 
 To add an entry for your Gramine service, run the `gramine-sgx-get-token` tool on the `.sig` file you built in the previous step as follows. (`gramine-sgx-get-token` is installed with [Gramine](https://github.com/gramineproject/gramine/).)
-
 
 ```bash
 gramine-sgx-get-token --sig hello.sig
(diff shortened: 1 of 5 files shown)


docs/docs/workflows/add-service.md
```diff
@@ -125,6 +125,8 @@ The environment variables have the following purposes.
 
 * `EDG_MARBLE_DNS_NAMES` is the list of DNS names and IPs the Coordinator will issue the Marble's certificate for.
 
+Optionally, you can set the `EDG_LOG_FORMAT` environment variable to `json` to enable JSON structured logs for the Marble's start up code.
```
thomasten, line 128:
```suggestion
Optionally, you can set the `EDG_LOG_FORMAT` environment variable to `json` to enable JSON-structured logs for the Marble's startup code.
```

injector/injector.go
```diff
@@ -244,7 +248,9 @@ func (m *Mutator) mutate(body []byte) ([]byte, error) {
 		})
 	}
 
-	return m.generateResponse(pod, admReviewReq, admReviewResponse, true, fmt.Sprintf("Mutation request for pod of marble type [%s] successful", marbleType))
+	msg := fmt.Sprintf("Mutation request for pod of marble type [%s] successful", marbleType)
+	m.log.Info(msg)
+	return m.generateResponse(pod, admReviewReq, admReviewResponse, true, msg)
```
thomasten, line 253:
```suggestion
	resp, err := m.generateResponse(pod, admReviewReq, admReviewResponse, true, msg)
	if err != nil {
		m.log.Error("Unable to mutate request: response generation failed", zap.Error(err))
		return nil, err
	}

	m.log.Info(msg)
	return resp, nil
```

marble/premain/premain.go
```diff
@@ -24,14 +24,27 @@ import (
...
 	"google.golang.org/grpc/credentials"
 )
 
+// LogJSON replaces Go's standard logger with [*zap.Logger] to log in JSON format.
+func LogJSON() (*zap.Logger, error) {
+	zapLogger, err := logging.New()
+	if err != nil {
+		return nil, err
+	}
+	defer zapLogger.Sync()
```
thomasten, line 43:
```suggestion
```
callers need to do this

thomasten (changes requested):
I can take care of ERT/EGo

injector/injector.go
```diff
@@ -244,7 +248,14 @@ func (m *Mutator) mutate(body []byte) ([]byte, error) {
 		})
 	}
 
-	return m.generateResponse(pod, admReviewReq, admReviewResponse, true, fmt.Sprintf("Mutation request for pod of marble type [%s] successful", marbleType))
+	msg := fmt.Sprintf("Mutation request for pod of marble type [%s] successful", marbleType)
+	resp, err := m.generateResponse(pod, admReviewReq, admReviewResponse, true, msg)
+	if err != nil {
+		return nil, err
```
thomasten, line 254:
```suggestion
		m.log.Error("Unable to mutate request: response generation failed", zap.Error(err))
		return nil, err
```
shouldn't we log the error here?

thomasten (approved):

PR #777, opened 2024-12-12
coordinator: fix equality checks for manifest properties

daniel-weisse:
two missing checks were preventing manifest updates from being properly applied if only these values changed

### Proposed changes
- fix missing check for `DisableSecretBinding` in `manifest.Marble.Equal`
- fix missing check for `AcceptedTCBStatuses` and `AcceptedAdvisories` in `quote.PackageProperties.Equal`

### DIFF
--- a/coordinator/manifest/manifest.go
+++ b/coordinator/manifest/manifest.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/edgelesssys/marblerun/coordinator/quote"
 	"github.com/edgelesssys/marblerun/coordinator/user"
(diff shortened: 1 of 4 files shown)


coordinator/manifest/manifest.go
```diff
@@ -98,22 +99,14 @@ func (m Marble) Equal(other Marble) bool {
 		return false
```
thomasten, line 99: nit: `if len(m.TLS) != len(other.TLS)` was also part of the slice comparison and can be removed

thomasten (approved):

PR #782, opened 2025-01-01
Fix Marble verification with Coordinator root certificate

thomasten:
External apps should be able to verify connections to Marbles by either the Coordinator root certificate or the Coordinator intermediate certificate ([docs](https://docs.edgeless.systems/marblerun/architecture/security#public-key-infrastructure-and-certificate-authority)). For this to work, the Marble's certificate chain must include the Coordinator intermediate certificate, not the Marble Root certificate. This chain is verifiable by any of (i) Coordinator Root, (ii) Coordinator Intermediate, (iii) Marble Root.

### Proposed changes
- Add an integration test that covers (i) and (ii) [(iii) already covered by multiple tests where test marble client connects to test marble server]
- Replace Marble Root with Coordinator Intermediate in Marble cert chain to fix (i)


coordinator/core/marbleapi.go
```diff
@@ -368,9 +369,13 @@ func customizeParameters(params manifest.Parameters, specialSecrets reservedSecr
...
 		return nil, fmt.Errorf("encoding Coordinator root CA: %w", err)
 	}
+	coordinatorIntermediatePem, err := manifest.EncodeSecretDataToPem(specialSecrets.CoordinatorIntermediate.Cert)
+	if err != nil {
+		return nil, fmt.Errorf("encoding Coordinator intermediate certificate: %w", err)
+	}
 
 	customParams.Env[marble.MarbleEnvironmentRootCA] = []byte(rootCaPem)
-	customParams.Env[marble.MarbleEnvironmentCertificateChain] = []byte(marbleCertPem + rootCaPem)
+	customParams.Env[marble.MarbleEnvironmentCertificateChain] = []byte(marbleCertPem + coordinatorIntermediatePem)
```
daniel-weisse, line 378: Won't this cause issues for some clients (e.g. openssl) because the intermediate certificate is not a terminating certificate?
Its also technically a breaking change.

Couldn't we add another environment variable that contains the certificate chain of leaf -> intermediate -> coordinator root

thomasten: I also thought about whether this is a breaking change, but came to the conclusion that it's a bug (only):
* It's uncommon to include the terminating certificate in the chain a TLS entity presents to the other side. Most servers present one chain including all but the root cert.
* Certification-path-building algorithms are supposed to find all paths leading to one of their known trusted certificates. So the other intermediate certificate in the chain should not affect verification if the Marble root cert is trusted.

So for reasonably good verifier implementations, the new chain should be accepted in all situations where the old chain was accepted plus in case the trust anchor is the Coordinator root cert only.

I manually tested openssl with
* Coordinator Root
* Intermediate
* Coordinator Root + Intermediate

as trust anchor. IIRC the problem back then was that openssl didn't accept a CA certificate as trusted if it's not self-signed.

I can't rule out the possibility that there are verifier implementations that would work with the old chain but don't work with the new chain. Users can apply workarounds in such situations (e.g., using only the Marble cert as chain or concatenating the two certs themselves).

Do you agree with all of the above or do you still have objections?

Do we have any openssl tests? If not, I can add some to the e2e test.

PR #785, opened 2025-01-07
coordinator: threshold manifest update support

daniel-weisse:
### Proposed changes
- Add a new manifest option `.Config.UpdateThreshold` to control the number of acknowledgements required to perform a full manifest update
  - This allows to define n users with update permission, but only require m users to acknowledge an update before its applied. Where m <= n
  - Can not be higher than the amount of users allowed to update the full manifest
  - If not set (0), all users have to acknowledge the update

### DIFF
--- a/coordinator/manifest/manifest.go
+++ b/coordinator/manifest/manifest.go
@@ -75,6 +75,9 @@ type Config struct {
 	SealMode string
 	// FeatureGates is a list of additional features to enable on the Coordinator.
 	FeatureGates []string
+	// UpdateThreshold is the amount of acknowledgements required to perform a multi party manifest update.
+	// If set to 0, all users with the update permission are required to acknowledge an update before it is applied.
+	UpdateThreshold uint
 }
 
 // Marble describes a service in the mesh that should be handled and verified by the Coordinator.
@@ -494,6 +497,19 @@ func (m Manifest) Check(zaplogger *zap.Logger) error {
 		default:
 			return fmt.Errorf("unknown type: %s for secret: %s", s.Type, name)
 		}
+	}
+
+	var manifestUpdaters int
+	for _, mrUser := range m.Users {
+		for _, roleName := range mrUser.Roles {
+			if m.Roles[roleName].ResourceType == "Manifest" && strings.ToLower(m.Roles[roleName].Actions[0]) == user.PermissionUpdateManifest {
+				manifestUpdaters++
+				break // Avoid counting the same user multiple times if they are assigned more than one role with update permission
+			}
+		}
+	}
+	if manifestUpdaters < int(m.Config.UpdateThreshold) {
+		return fmt.Errorf("not enough users with manifest update permissions (%d) to meet the threshold of %d", manifestUpdaters, m.Config.UpdateThreshold)
 	}
 
 	switch m.Config.SealMode {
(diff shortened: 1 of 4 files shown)


coordinator/manifest/manifest.go
```diff
@@ -496,6 +499,19 @@ func (m Manifest) Check(zaplogger *zap.Logger) error {
...
+	var manifestUpdaters int
+	for _, mrUser := range m.Users {
+		for _, roleName := range mrUser.Roles {
+			if m.Roles[roleName].ResourceType == "Manifest" && strings.ToLower(m.Roles[roleName].Actions[0]) == user.PermissionUpdateManifest {
+				manifestUpdaters++
+				break // Avoid counting the same user multiple times if they are assigned more than one role with update permission
+			}
+		}
+	}
+	if manifestUpdaters < int(m.Config.UpdateThreshold) {
```
thomasten, line 511:
```suggestion
	var manifestUpdaters uint
	for _, mrUser := range m.Users {
		for _, roleName := range mrUser.Roles {
			if m.Roles[roleName].ResourceType == "Manifest" && strings.ToLower(m.Roles[roleName].Actions[0]) == user.PermissionUpdateManifest {
				manifestUpdaters++
				break // Avoid counting the same user multiple times if they are assigned more than one role with update permission
			}
		}
	}
	if manifestUpdaters < m.Config.UpdateThreshold {
```
nit: change type to avoid cast

coordinator/manifest/manifest.go
```diff
@@ -496,6 +499,19 @@ func (m Manifest) Check(zaplogger *zap.Logger) error {
 		}
 	}
 
+	var manifestUpdaters int
+	for _, mrUser := range m.Users {
+		for _, roleName := range mrUser.Roles {
+			if m.Roles[roleName].ResourceType == "Manifest" && strings.ToLower(m.Roles[roleName].Actions[0]) == user.PermissionUpdateManifest {
```
thomasten, line 505: Actions can be empty and cause a panic here

coordinator/manifest/manifest_test.go
```diff
@@ -380,11 +381,39 @@ func TestManifestCheck(t *testing.T) {
...
+	adminUser2 := adminUser
+	adminUser2.Certificate = strings.ReplaceAll(adminUser.Certificate, "q", "b")
+	manifest.Users["admin2"] = adminUser2
+	err = manifest.Check(log)
+	assert.NoError(err, "update threshold lower than allowed users should be accepted")
+
+	manifest.Config.UpdateThreshold = 3
+	err = manifest.Check(log)
+	assert.Error(err, "update threshold higher than allowed users should be rejected")
+	manifest.Config.UpdateThreshold = 0
```
thomasten, line 410: IMO the additions would satisfy refactoring this test to table-driven

coordinator/manifest/manifest.go
```diff
@@ -496,6 +499,19 @@ func (m Manifest) Check(zaplogger *zap.Logger) error {
...
+				manifestUpdaters++
+				break // Avoid counting the same user multiple times if they are assigned more than one role with update permission
```
thomasten, line 507: add a test case that it fails if you remove the `break`

thomasten (changes requested):

thomasten (approved):

PR #793, opened 2025-01-28
charts: allow configuration of Coordinator PVC size through helm values file

Nabsku:
### Proposed changes
- Some storage backends such as NetApp Trident technically only allow volumes bigger than 20Mi. We instead deploy a separate claim now that we use in the marblerun chart. As a little UX improvement I propose that we allow setting the pvc size via the values.yaml file


daniel-weisse (approved):
Thanks, looks good

PR #808, opened 2025-03-13
cli: report accurate number of missing acknowledgements and users on update

daniel-weisse:
### Proposed changes
- Report the actual number of missing acknowledgements on `marblerun manifest update acknowledge`
- Print a list of all users who may still acknowledge the update
- Print the number of expected acknowledgements on `marblerun manifest update apply` in case of full manifest update

### Additional info
- This PR introduces a breaking change in the Go API for `ManifestUpdateAcknowledge`. The function now returns a list of users who have not acknowledged the update as well as the number of users that still need to send an acknowledgement for the update to be applied
- This PR introduces a breaking change in the Go API for `ManifestUpdateApply`. The function now returns a list of a users who may acknowledge the update, as well as the number of missing acknowledgements for the update to be applied (in case of full manifest updates with more than 1 user)

### DIFF
--- a/cli/internal/cmd/manifestUpdate.go
+++ b/cli/internal/cmd/manifestUpdate.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/edgelesssys/marblerun/api"
 	"github.com/edgelesssys/marblerun/cli/internal/certcache"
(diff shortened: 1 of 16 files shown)


cli/internal/cmd/manifestUpdate.go
thomasten, line 1: Do we need to update the docs for any of the changes? I remember that we at least wanted to clarify that UpdateThreshold includes the initial `update apply`

daniel-weisse: Updated the documentation

thomasten (approved):

PR #850, opened 2025-09-18
coordinator: seal with 32 byte key

daniel-weisse:
### Proposed changes
- Seal encryption key using `SealWithProductKey256`
  - Unsealing happens using `Unseal256` or `Unseal` if unsealing with the 32 byte key fails as a fallback
- Seal data using a 32 byte key
  - A 32 byte sealing key is only generated for new MarbleRun deployments. Deployments upgrading from a previous MarbleRun version will continue using a 16 byte key for data sealing

### Additional info
- AB#6112

### DIFF
--- a/coordinator/recovery/recovery.go
+++ b/coordinator/recovery/recovery.go
@@ -43,7 +43,7 @@ func ParseRSAPublicKeyFromPEM(pemContent string) (*rsa.PublicKey, error) {
 }
 
 func generateRandomKey() ([]byte, error) {
-	generatedValue := make([]byte, 16)
+	generatedValue := make([]byte, 32)
 	_, err := rand.Read(generatedValue)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


coordinator/recovery/recovery.go
```diff
@@ -43,7 +43,7 @@ func ParseRSAPublicKeyFromPEM(pemContent string) (*rsa.PublicKey, error) {
 }
 
 func generateRandomKey() ([]byte, error) {
-	generatedValue := make([]byte, 16)
+	generatedValue := make([]byte, 32)
```
thomasten, line 46: maybeDecryptRecoveryKey in `cli/internal/cmd/recover.go` and DEK len in `docs/docs/architecture/security.md` needs to be adapted to this. (No need to mention existing deployments.)

thomasten (changes requested):

coordinator/recovery/recovery.go
daniel-weisse: Good catch
Didnt notice during testing, since I never manually decrypted the recovery keys

PR #851, opened 2025-09-22
cli: allow specifying custom qcnl config when installing MarbleRun

daniel-weisse:
### Proposed changes
- Add a new flag to to `marblerun install`: `--dcap-qcnl-config-file`
  - Allows specifying the path to a qcnl config, which will be added to the Helm chart instead of using the default values in `--dcap-pccs-url` and `--dcap-secure-cert`
- The Helm chart similarly received a new settable value: `dcap.qcnlConfig`
  - Allows defining a qcnl config inline as a string
  - Alternatively the value may be set by reading a file using Helm's `--set-file` flag:
      ```
      helm install -n marblerun ./charts --set-file dcap.qcnlConfig=/path/to/sgx_qcnl.conf
      ```
- Update the default qcnl generated by the CLI and the Helm chart to use the newer JSON formatted config

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -62,8 +62,9 @@ their default values.
 | `marbleInjector.namespaceSelector`           | object         | NamespaceSelector to trigger marble-injector mutation, See the [K8S documentation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector) for more information | `{}` |
 | `nodeSelector`                               | object         | NodeSelector section, See the [K8S documentation](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector) for more information | `{"beta.kubernetes.io/os": "linux"}` |
 | `tolerations`                                | object         | Tolerations section, See the [K8S documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) for more information | `{key:"sgx.intel.com/epc",operator:"Exists",effect:"NoSchedule"}` |
-| `dcap.pccsUrl`                               | string         | URL of the PCCS | `"https://global.acccache.azure.net/sgx/certification/v4/"` |
-| `dcap.useSecureCert`                         | string         | Whether or not the TLS certificate of the PCCS should be verified | `"TRUE"` |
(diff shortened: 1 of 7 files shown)


charts/README.md
```diff
@@ -62,8 +62,9 @@ their default values.
...
 | `nodeSelector`                               | object         | NodeSelector section, See the [K8S documentation](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector) for more information | `{"beta.kubernetes.io/os": "linux"}` |
 | `tolerations`                                | object         | Tolerations section, See the [K8S documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) for more information | `{key:"sgx.intel.com/epc",operator:"Exists",effect:"NoSchedule"}` |
-| `dcap.pccsUrl`                               | string         | URL of the PCCS | `"https://global.acccache.azure.net/sgx/certification/v4/"` |
-| `dcap.useSecureCert`                         | string         | Whether or not the TLS certificate of the PCCS should be verified | `"TRUE"` |
+| `dcap.qcnlConfig`                            | string         | Inline defined QCNL configuration. If set, this configuration is used instead of the default one or the one provided via `dcap.qcnlConfigFile` | `""` |
```
thomasten, line 65: I guess qcnlConfigFile was an earlier approach to solve this that you discarded?

charts/templates/sgx_qcnl.yaml
```diff
@@ -13,7 +13,14 @@ metadata:
...
+  {{- if .Values.dcap.qcnlConfig }}
+  sgx_default_qcnl.conf: |-
+{{ .Values.dcap.qcnlConfig | indent 4 }}
+  {{- else }}
   sgx_default_qcnl.conf: |
-    PCCS_URL={{ .Values.dcap.pccsUrl }}
-    USE_SECURE_CERT={{ .Values.dcap.useSecureCert }}
+    {
+      "pccs_url": "{{ .Values.dcap.pccsUrl }}",
+      "use_secure_cert": "{{ .Values.dcap.useSecureCert }}"
```
thomasten, line 23: I think this must be a json bool, not, e.g., `"TRUE"`.

